PR #50, opened 2018-07-27
Remove enumerated type, use Enum decorator instead

tshemsedinov:


tshemsedinov:
Landed in 228e1ce and 228e1ce

PR #54, opened 2018-07-28
Update and move schemas

tshemsedinov:


taxonomy/domains.schema
```diff
@@ -9,16 +9,46 @@
   URL: { type: 'string', length: 2000, control: 'InputBox' },
   IP: { type: 'string', length: 15, control: 'InputBox' },
   SID: { type: 'string', length: 64 },
-  Login: { type: 'string', length: 32 },
+  Login: { type: 'string', length: 32,  },
```
lundibundi: Is this trailing comma intentional?

taxonomy/domains.schema
```diff
@@ -9,16 +9,46 @@
...
+    A: 'Audit'
+  }),
   Text: { type: 'string', control: 'Multiline' },
+  Values: { type: 'object' },
   JSON: { type: 'string', control: 'Multiline' },
   HTML: { type: 'string', control: 'Multiline' },
   Year: { type: 'number', length: 4, control: 'InputBox' },
-  Month: Enum('January', 'February', 'March', 'April', 'May', 'June', 'July', 'August', 'September', 'October', 'November', 'December'),
-  Day: Enum('Sunday', 'Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday'),
+  Month: Enum({
```
lundibundi: Maybe it's worth adding `startOffset` option to `enumArray` to avoid typing such enums by hand?

tshemsedinov: thanks

taxonomy/domains.schema
tshemsedinov: thanks

taxonomy/domains.schema
tshemsedinov: @lundibundi I can't find good syntax for this and this will be storage data type, so we can't rename keys implicitly. Also we have multiple `Enum` use cases, `Enum(10, 20, 25)` for numeric constants stored as number, `Enum('str1', 'str2', 'str3')` for strings stored as strings, `Enum({ key1: value1, key2: value2, key3: value3 })` stored as string (because key can be string). For strings we can convert to numbers but sometimes we may want to use strings. If we want to store keys as numbers we need special syntax. If we want to store enumerated strings as numbers we also need special syntax or we can always start enumeration from 1. For example `Enum({ 1: value1, 2: value2, 3: value3 }, 'number')` and `Enum(['str1', 'str2', 'str3'], 'number')` or `Enum({ 1: value1, 2: value2, 3: value3 }).store('number')` and `Enum('str1', 'str2', 'str3').store('number')`. But I don't like any of mentioned cases.

tshemsedinov:
Landed in 8e56c0d and f80d16b

taxonomy/domains.schema
lundibundi: @tshemsedinov I thought of something like
`Enum({ values: ['abc', 'aa'], startOffset: 1 })` or
`Enum(['a', b'], { startOffset: 1 })`
First will work as it's not allowed to have different types of variables in an enum. The second just has additional `optinos` argument which will allow for futher customization if needed.
What do you think? I'm personally in favor of the second one as it's much more flexible and less obtrusive. Though it may block `enum of 2 objects` but I'm not sure if it's even allowed.

PR #60, opened 2018-08-02
Handle errors, don't instantiate if validation fails

tshemsedinov:


tshemsedinov:
Landed in 9425953 and b49633b

PR #67, opened 2018-08-18
Refactor schemas

tshemsedinov:


tshemsedinov:
Landed in ad3d69c, 13cf85a and 532b83f

lundibundi:
Hm, so now even in `metashema` all schemas are in `scheme` folder.
Though could you please leave the PRs open for at least a couple of hours for other to be able to take a look?
Also, s/Memove Details decorator/Remove Details decorator/ in commit message.

lundibundi:
@tshemsedinov you can just use npm filesystem dependencies, you can just issue `npm i -S ../path/to/packet/root` and you will be able to test packages locally without issuing a new version.

Though I understand your arguments and agree that it may be hard to test some changes locally. But it'd still be better to wait at least for a few hours before landing.

PR #72, opened 2018-08-21
Simplify relation decorators for globalstorage

tshemsedinov:


tshemsedinov:
Landed in 8e14729, 3a252bd and 1a7d894

lib/decorators.js
```diff
@@ -14,10 +14,20 @@ class Decorator {
...
 }
 
-class Many extends Decorator {}
-class Parent extends Decorator {}
-class Master extends Decorator {}
-class Include extends Decorator {}
+class RelationDecorator extends Decorator {
+  constructor(def) {
+    const defType = typeof def;
+    if (defType === 'string') {
```
lundibundi: Why not one-liner? (`typeof def === 'string`)

schemas/CategoryField.schema
```diff
@@ -5,7 +5,9 @@
...
   comment: { domain: 'Text' },
 
-  Check: Validate(data => (
-    +!!data.domain + +!!data.category === 1
-  ))
+  Check: Validate(data => {
+    const isDomain = !!data.domain;
+    const isCategory = !!data.category;
+    return isDomain ? !isCategory : isCategory;
+  })
```
lundibundi: `return data.domain ? !data.category : data.category`?

schemas/DatabaseField.schema
```diff
@@ -2,13 +2,14 @@
   domain: { domain: 'Nomen' },
   category: { domain: 'Nomen' },
   required: { domain: 'Logical' },
-  unique: { domain: 'Logical' },
-  index: { domain: 'Logical' },
-  default: { domain: 'Nomen' },
+  unique: { domain: 'Logical' }, // single field unique key
+  index: { domain: 'Logical' }, // single field index
+  default: { domain: 'Nomen' }, // default value to be casted to domain
```
lundibundi: What's the meaning of `casted` here? Maybe `used`?

schemas/DatabaseField.schema
```diff
@@ -31,14 +32,14 @@
 
     // Validate calculated field
     if (data.expression) {
-      return allowed(['domain', 'category', 'control', 'comment']);
+      return allowed(['domain', 'display', 'category', 'control', 'comment']);
```
lundibundi: :thinking: shouldn't it be `allow(..)`?

schemas/DatabaseField.schema
```diff
@@ -31,14 +32,14 @@
```
lundibundi: Also, you can convert the above lambda to
```js
const allow = allowed => fields.every(f => allowed.includes(f));
```

schemas/CategoryField.schema
tshemsedinov: `Check: Validate(data => data.domain ? !data.category : !!data.category)` because function should always return Boolean

schemas/DatabaseField.schema
tshemsedinov: I mean type casting

schemas/DatabaseField.schema
tshemsedinov: We have `allowed(allow)` but you are right, it would be better to have `allow(allowed)`

schemas/CategoryField.schema
lundibundi: That's good too.

lundibundi:
Should I open a PR to fix those or you will fix them?

tshemsedinov:
@lundibundi I'll fix that, thanks

tshemsedinov:
Fixed in https://github.com/metarhia/metaschema/pull/74

PR #74, opened 2018-08-22
Fix bugs and code style in schemas

tshemsedinov:


schemas/DatabaseField.schema
```diff
@@ -23,41 +23,36 @@
     const fields = Object.keys(data);
 
     // Check allowed fields
-    const allow = allowed => {
-      for (const field of fields) {
-        if (!allowed.includes(field)) return false;
-      }
-      return true;
-    };
+    const allow = allowed => fields.every(f => allowed.includes(f));
```
lundibundi: What do you think of using Set for `allowed` here?
Then this should be cleaner and maybe a bit faster.

Also, I'd suggest to rename this method to `checkFields`, `chechFieldConstraints` or something. Because `allow` suggest that this method somehow 'allows' fields provided.

lundibundi:
Also, `default: { domain: 'Nomen' }, // default value to be casted to domain` could you clarify. From your comment I understand that this value will be casted to some-type (what type)?

tshemsedinov:
@lundibundi here I use `default: { domain: 'Nomen' }` just to define string field because we can't use types in db schemas, only domains but we have no more suitable domain.

schemas/DatabaseField.schema
lundibundi: Using array is fine with me too. Though what about the renaming bit?

tshemsedinov: mb `allowFields` because it allows subset of record fields:
```js
const fits = allowFields(['category', 'comment']);
if (!fits) return false;
```

tshemsedinov:
@lundibundi I am going to define special domain for such values of unknown type to casted in runtime into certain type using metadata: `default: { domain: 'Literal' }`

tshemsedinov:
Landed in d02bd0d and cc058cd

PR #82, opened 2018-08-26
Update Domain.schema file

lundibundi:
* Rename to DomainField.schema to be consistent with other such files
  (CategoryField.schema and DatabaseField.schema)
* Add 'class' property (it is used in domains.schema but was missing
  from this file).

### DIFF
--- a/schemas/metaschema/DomainField.schema
+++ b/schemas/metaschema/DomainField.schema
@@ -1,5 +1,6 @@
 {
   type: Enum('string', 'number', 'boolean', 'function', 'object', 'symbol'),
+  class: { type: 'string' },
   control: { type: 'string' },
   length: { type: 'number' },
   unit: { type: 'string' },


schemas/metaschema/DomainField.schema
```diff
@@ -1,5 +1,6 @@
 {
   type: Enum('string', 'number', 'boolean', 'function', 'object', 'symbol'),
+  class: { type: 'string' },
```
tshemsedinov: Why do we need `class` here and this may give a conflict with keyword.

lundibundi: I meant `DomainField` as in field in `domains.schema` object, so it's not the Domain field, it's a possible properties of Domain definition.

I've added `class` because it's used in `domain.schema` (https://github.com/metarhia/metaschema/blob/master/schemas/metaschema/domains.schema#L18)

tshemsedinov:
Landed in 7f41d49932ec50c7a06b08b2f61658f381f73bcb

PR #86, opened 2018-08-27
Decorators Bitmask, Local, History, Master

tshemsedinov:
- Added `Bitmask`, `Local`, `History`
- Changed `Master`: now it will set `required: true` by default


tshemsedinov:
Landed in 40ce903

PR #91, opened 2018-08-29
Fix and improve schemas

tshemsedinov:
- Add Memory decorator
- Add comment attribute
- Add SHA256 domain

### DIFF
--- a/schemas/metaschema/domains.schema
+++ b/schemas/metaschema/domains.schema
@@ -13,6 +13,7 @@
   Login: { type: 'string', length: 32 },
   Password: { type: 'string', length: 32, control: 'InputBox' },
   PasswordHash: { type: 'string', length: 128 },
+  SHA256: { type: 'string', length: 32 },
   Salt: { type: 'string', length: 32 },
   Text: { type: 'string', control: 'Multiline' },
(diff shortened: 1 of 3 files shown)


tshemsedinov:
Landed in 8536de1

schemas/metaschema/domains.schema
```diff
@@ -13,6 +13,7 @@
   Login: { type: 'string', length: 32 },
   Password: { type: 'string', length: 32, control: 'InputBox' },
   PasswordHash: { type: 'string', length: 128 },
+  SHA256: { type: 'string', length: 32 },
```
belochub: I think we should preferably use SHA-512/256 since it is somewhat faster on 64-bit processors.

Also, using type `string` here may lead to some problems because we actually store 32 bytes of 'random' data, and using strings to store the data may lead to them having invalid Unicode code points inside them which may be replaced with 'REPLACEMENT CHARACTER' (U+FFFD) by some of the Unicode parser/serializers and it may lead to losing some data in the process. We probably should replace the type here with `Uint8Array` or something similar. Or we could use base64 serialization to store this binary data in a string, but it will take up 44 bytes instead of 32 in that case.

Issue #105, opened 2018-09-14
Missing domains


nechaido:
In master branch of globalstorage `metaschema.domains` is missing many domains after loading `'schemas/system'`, the only ones present are:
```
Map {
  'IdStatus' => Enum { values: [ 'Prealloc', 'Init', 'Actual', 'Historical' ] },
  'StorageKind' => Enum { values: [ 'Master', 'Cache', 'Backup', 'Replica' ] },
  'CategoryRealm' => Enum { values: [ 'Global', 'System', 'Local' ] },
  'CategoryFamily' => Enum {
    values:
     [ 'Registry',
       'Dictionary',
       'Table',
       'Log',
       'View',
       'History',
       'Memory' ] },
  'UserAccess' => Enum {
    values: [ 'Read', 'Insert', 'Update', 'Delete', 'Execute', 'Audit' ] },
  'AccessFlags' => Flags {
    values: [ 'Read', 'Insert', 'Update', 'Delete', 'Execute', 'Audit' ],
    enum: 'UserAccess' },
  'Compression' => Enum { values: [ 'None', 'GZIP', 'ZIP' ] },
  'SessionToken' => { type: 'string', length: 64 },
  'MediaType' => Enum {
    values:
     [ 'text/plain',
       'text/csv',
       'image/png',
       'image/gif',
       'image/jpeg',
       'image/svg+xml',
       'image/tiff',
       'application/pdf',
       'application/rtf',
       'application/zip',
       'application/json' ] },
  'SystemType' => Enum { values: [ 'Global', 'Autonomous', 'Standalone' ] },
  'ServerType' => Enum { values: [ 'Root', 'Server', 'Backup', 'Reserve' ] },
  'NodeType' => Enum { values: [ 'Master', 'Worker', 'Task' ] },
  'NodeStatus' => Enum { values: [ 'Init', 'Active', 'Pause', 'Restart', 'Hang' ] } }
```
Code to reproduce:
```js
'use strict';
const metaschema = require('metaschema');
metaschema.load('schemas/system', (error, schema) => {
  if (error) {
    console.error(error);
    return;
  }

  console.log(metaschema.domains);
});

```

tshemsedinov:
Here are all domains from `schemas/system/domains.schema` @nechaido

PR #106, opened 2018-09-19
Load default domains (metaschema) at require time

lundibundi:
Fixes: https://github.com/metarhia/metaschema/issues/105

### DIFF
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -3,13 +3,12 @@
 const vm = require('vm');
 const fs = require('fs');
 const path = require('path');
-const { basename } = require('path');
 const metasync = require('metasync');
 
 const { parseSignature } = require('./introspection.js');
 const decorators = require('./decorators.js');
 const moduleDir = path.dirname(__dirname);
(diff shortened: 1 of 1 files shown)


lundibundi:
@tshemsedinov review this again please as I have replaced async load with a synchronous load as using async functions may introduce race conditions here.
Also, I've added a second commit to remove manual separator ('/') in the file. This is to be landed as 2 commits.

lib/schema.js
```diff
@@ -3,13 +3,13 @@
 const vm = require('vm');
 const fs = require('fs');
 const path = require('path');
-const { basename } = require('path');
 const metasync = require('metasync');
 
 const { parseSignature } = require('./introspection.js');
```
SemenchenkoVitaliy: I think `'/'` here should be changed as well

belochub:
Landed in https://github.com/metarhia/metaschema/commit/ebcecfbc88e252edfa3e59523b93766f1f25f9e7 and https://github.com/metarhia/metaschema/commit/231a08c421584cea0f68642f1f10bc02cb28f367.

PR #107, opened 2018-09-21
Introspect refactor to support new comments style

SemenchenkoVitaliy:


SemenchenkoVitaliy:
@tshemsedinov Is there any place, where `introspect` and `parseSignature` or `guard`, which depends on it, is used? Because there is only one way to remain backwards-compatible - read files content inside parser and I'm not a fan of this idea

lib/introspection.js
```diff
@@ -71,28 +69,120 @@ const parseSignature = (
...
+  const comments = [];
+
+  for (let i = lines.length - 2; i >= 0; i--) {
+    if (lines[i].indexOf('// ') !== 0) break;
+    comments.unshift(lines[i].slice(3));
+  }
+  return comments;
+};
+
+const parseComments = (lines) => {
```
tshemsedinov: We need description and use examples for new implementation, also: change `(lines) => {` to `lines => {`

lib/introspection.js
```diff
@@ -71,28 +69,120 @@ const parseSignature = (
   return signature;
 };
 
-const introspect = (
-  // Introspect interface
-  namespace // hash of interfaces
-  // Returns: hash of hash of record, { method, title, parameters }
-) => {
+const parseFunction = (fn, text) => {
+  const fnData = fn + '';
```
tshemsedinov: `fn.toString()`

lib/introspection.js
```diff
@@ -71,28 +69,120 @@ const parseSignature = (
...
+      sig.comments.push(obj);
+    }
+  }
+  return sig;
+};
+
+// Introspect interface
+//   namespace - hash of interfaces
+// Returns: hash of hash of record, { method, title, parameters }
+const introspectOld = (namespace, text) => {
```
tshemsedinov: Why do we need old versions?

lundibundi: @tshemsedinov that's related to https://github.com/metarhia/metaschema/pull/107#issuecomment-423511723. We still have usages of old introspect in `guard` function (it gets type from the comment and checks for it).
I'd suggest to change the contract of guard to pass the argument types as an argument instead of getting them from the code-comment.
Otherwise we would have to keep in memory all of the files with definitions in order to get the code-commments OR pass filename to this function (and then require + source parse it) OR pass source of the file to this function (as it is now). Though I prefer signature change as all of the other variants are more like hacks and not the actual implementation.

lib/schema.js
```diff
@@ -263,12 +260,15 @@ const validate = (
...
-) => {
+// Validate
+//   category - Category name
+//   instance - Instance to validate
+//   result - object, default: {valid: true, errors: []}
+//     valid - boolean
+//     errors - array of errors
+// Returns: object, mutated result
+const validateFields = (category, instance,
+  result = { valid: true, errors: [] }) => {
```
tshemsedinov:
```
const validateFields = (
  category, instance,
  result = { valid: true, errors: [] }
) => {
```

lib/schema.js
```diff
@@ -198,29 +195,29 @@ const factorify = (
...
-  // Build schemas list (directory)
-  schemas // object, key: schema
-  // Returns: function
-) => {
+// Build schemas list (directory)
+//   schemas - object, key: schema
+// Returns: function, with parameters:
+//   name - category name
+//   ...args - fields to create instance
+const build = (schemas) => {
```
tshemsedinov: `schemas => {`

lib/schema.js
lundibundi: +1 but put every argument on a separate line. wdyt?

lib/introspection.js
```diff
@@ -71,28 +76,140 @@ const parseSignature = (
...
+      if (types.length === 1) {
+        if (!ALL_TYPES.find(s => types[0].startsWith(s))) {
+          comment = rest;
+          type = '';
+        } else {
+          type = types[0];
+        }
+      } else {
+        type = [];
+        console.log(types);
```
nechaido: Debug symbol.

lib/introspection.js
```diff
@@ -71,28 +76,140 @@ const parseSignature = (
...
+};
+
+
+// Introspect interface
+//   namespace - <object>, hash of interfaces
+//   text - <string>, data to parse
+// Returns: <object>, hash of hash of record, { method, title, parameters }
+const introspect = (namespace, text) => {
+  const inventory = {};
+  for (const name in namespace) {
```
nechaido: I'd prefer to use `Map` for `namespace` if it is not used anywhere else.

lib/introspection.js
```diff
@@ -71,28 +76,140 @@ const parseSignature = (
...
+    const iface = namespace[name];
+    const methods = {};
```
nechaido: Ditto.

nechaido: It looks like `methods` may store info about `Class`-es, in that case it would be better to rename it to something like `entities` or `comments`, though I am open to a better naming proposal.

tshemsedinov:
@SemenchenkoVitaliy please rebase with master

test/example.js
```diff
@@ -1,20 +1,28 @@
...
-  // Returns: string, comment
-  // Hint: message
-  // Example: methodName(5, [2, 3], 'f')
-  // Result: '52,3f'
-) => {
+// Method caption
+// First line description
+// second line of multiline method description
+//   a1 - <number>, comment
+//     multiline comment, something
```
lundibundi: Should be aligned with the above (here and below).

test/example.js
```diff
@@ -1,20 +1,28 @@
...
+//     for the first argument
+//   a2 - <string[]>, comments
+//   a3 - comment
+//   fn - <function>, fn(a, b, c), description
+//     a - <object>, also description
+//     b - <boolean> | <number>
+//     c - <array>, comment
+// Returns: <string>, comment
+// Hint: message
```
lundibundi: Why do we need `hint`?
I mean, why don't we just put anything that might go in `hint` directly to the description up top?

SemenchenkoVitaliy: I don't know either though I suspect there can be something not related to usage like information about how internals work. If someone comes up with a better explanation, I'd be glad to hear it. As for now I think we should leave it here.

lib/introspection.js
```diff
@@ -1,13 +1,15 @@
 'use strict';
 
 const common = require('metarhia-common');
-const { ALL_TYPES } = require('./types');
+const ALL_TYPES = [
```
lundibundi: Why not a Set?

lib/introspection.js
```diff
@@ -79,17 +79,21 @@ const parseComments = lines => {
       let [type, comment] = common.section(rest, ', ');
       const types = type.split(' | ').map(s => s.slice(1, s.length - 1));
       if (types.length === 1) {
-        if (!ALL_TYPES.find(s => types[0].startsWith(s))) {
+        type = types[0];
+        console.log(types[0]);
```
lundibundi: debug output?

lib/introspection.js
```diff
@@ -79,17 +79,21 @@ const parseComments = lines => {
...
+        if (!(ALL_TYPES.has(type) ||
+            (ALL_TYPES.has(type.slice(0, type.length - 2)) &&
```
lundibundi: You can omit these parens I think.

SemenchenkoVitaliy: The closing parenthesis was in the wrong place. Thanks, though.

lundibundi: I mean. this can be:
```js
const arrType = type.slice(0, type.length - 2)
if (!ALL_TYPES.has(type) &&
   (!ALL_TYPES.has(arrType) || type.slice(type.length - 2) !== '[]')) {
```

lib/introspection.js
```diff
@@ -1,91 +1,158 @@
...
-        signature.description += line;
       } else {
-        const par = last(signature.parameters);
-        par.comment += '\n' + line;
+        type = [];
+        const notValidTypes = [];
+        types.forEach(s => {
+          if (ALL_TYPES.has(s) ||
+            (ALL_TYPES.has(s.slice(0, s.length - 2) &&
+              s.slice(s.length - 2) === '[]'))) {
```
lundibundi:
```js
const arrType = type.slice(0, type.length - 2)
if (ALL_TYPES.has(type) ||
   (ALL_TYPES.has(arrType) && type.slice(type.length - 2) !== '[]')) {
```
(also parens around `(ALL_TYPES.has(arrType) && type.slice(type.length - 2) !== '[]')` can be omitted if you want)

bin/cli.js
```diff
@@ -3,18 +3,25 @@
 
 const fs = require('fs');
 const path = require('path');
-
```
belochub: There is no need to remove this line, it was separating node.js module requires from npm module requires.

lib/generator.js
```diff
@@ -1,38 +1,55 @@
 'use strict';
 
+const common = require('metarhia-common');
+
 // Generate md from interfaces inventory
-//   inventory - hash of hash of record, { method, title, parameters }
-// Returns: string, md document
+//   inventory - <object>, hash of map of records, { method, title, parameters }
```
belochub:
```suggestion
//   inventory - <Object>, hash of map of records, { method, title, parameters }
```

lib/introspection.js
```diff
@@ -1,91 +1,160 @@
...
+const NAMED_LINES = [
+  'Returns:', // returned value
+  'Example:', // example of usage
+  'Result:', // result of usage, should be used with `Example`
+  'Throws:', // thrown errors
+  'Deprecated', // reasons for deprecation or previous names
+];
+
+// Search `text` for `fn` occurrences
+//   fn - <function>, to be searched
```
belochub:
```suggestion
//   fn - <Function>, to be searched
```

lib/introspection.js
```diff
@@ -1,91 +1,160 @@
...
+  return comments.map(line => line.trimRight()).filter(line => line.length);
+};
+
+const isNamedLine = line => NAMED_LINES.find(s => line.startsWith(s));
+
+const splitAtIndex = (line, index) => [line.slice(0, index), line.slice(index)];
+
+// Parse comments related to function
+//   lines - <string[]>
+// Returns: <object>, function signature
```
belochub:
```suggestion
// Returns: <Object>, function signature
```

lundibundi:
As I mentioned in a prev PR/comments, could we make the comma after argument type optional?
i.e.
`arg - <type> description` === `arg - <type>, description`

lib/introspection.js
```diff
@@ -78,7 +78,10 @@ const parseComments = lines => {
     const [name, rest] = common.section(arg, ' - ');
 
     if (rest !== '') {
-      let [type, comment] = common.section(rest, ', ');
+
+      let [type, comment] = splitAtIndex(rest, rest.lastIndexOf('>') + 1);
```
lundibundi: Wouldn't this fail in such case
`arg - <string> must have length > 13`
?

tshemsedinov:
- Please remove dash `arg - <type> description` -> `arg <type> description`
- What is solution for all scalar types? For example `id = x => x` will be described as `x <string> | <number> | <boolean> description`, we may optimize it to `x <scalar> description`

lundibundi:
@tshemsedinov 
* dash was decided to be optional AFAIU
* I think it'll be better to avoid such abbreviations that may not be known to some people and/or may be misunderstood as there is no such thing as 'scalar' in the spec AFAIK.

tshemsedinov:
@lundibundi `primitive` will be fine

SemenchenkoVitaliy:
I'd like to know whether `introspection` will be used in other files as it was i.e. whether we need backwards compatibility because now it if the argument has only one type it will be stored as a string, but if it has multiple types they will be stored as an array of strings. I didn't like that solution but back then there was `guard`  which is not present here anymore and it doesn't seem to be any functions that rely on `introspect`. With that said I want to propose to fix this and store all types as an array of strings.

bin/cli.js
```diff
@@ -8,13 +8,21 @@ const common = require('metarhia-common');
 
 const metaschema = require('..');
 
-const cwd = process.cwd();
-const apiFile = path.resolve(cwd, process.argv[2]);
+const apiFile = path.resolve(process.cwd(), process.argv[2]);
```
lundibundi: Wouldn't this be affected by the same issue as in browser-runner? (Shouldn't we search for project root dir?)

bin/cli.js
```diff
@@ -8,13 +8,21 @@ const common = require('metarhia-common');
...
 
 try {
   const imports = require(apiFile);
   const namespace = path.basename(apiFile, '.js');
-  const inventory = metaschema.introspect({ [namespace]: imports });
+  let data = '';
+  for (const module in require.cache) {
+    if (!module.includes('node_modules') && module !== __filename &&
+      fs.existsSync(module)) {
```
lundibundi:
```js
if (!module.includes('node_modules') &&
    module !== __filename &&
    fs.existsSync(module)) {
```
?

bin/cli.js
```diff
@@ -8,13 +8,21 @@ const common = require('metarhia-common');
...
+      data += fs.readFileSync(module, 'utf8') + '\n';
+    }
+  }
+
+  const inventory = metaschema.introspect(new Map([[namespace, imports]]),
+    data);
```
lundibundi:
```js
  const inventory = metaschema.introspect(
    new Map([[namespace, imports]]), data
  );
```
OR
```js
  const inventory =
    metaschema.introspect(new Map([[namespace, imports]]), data);
``` 
?

lib/generator.js
```diff
@@ -1,38 +1,159 @@
 'use strict';
 
+const common = require('metarhia-common');
+
+const structures = [
```
lundibundi: These should be `primitives` I think (https://developer.mozilla.org/en-US/docs/Glossary/Primitive).

lib/generator.js
```diff
@@ -1,38 +1,159 @@
...
+  'boolean',
+  'null',
+  'undefined',
+  'number',
+  'string',
+  'symbol',
+];
+
+const gObjects = [
```
lundibundi: What is `gObjects` and why not just `objects`?

Edit: I assume this is due to the naming in MDN though I think it'd be better to use full name `globalObjects`.

lib/generator.js
```diff
@@ -1,38 +1,159 @@
...
+links = links.map(link => '[`<' + link[0] + '>`]: ' + link[1]);
+
+const wrapType = type => {
+  if (types.includes(type)) {
+    return '[`<' + type + '>`][]';
+  } else if (type.startsWith('[')) {
+    type = type.slice(1, type.length - 1)
+      .replace(/ /, '')
+      .split(',')
+      .map(s => s.slice(1, s.length - 1))
```
lundibundi: This will probably fail for smth like `[[<A>, <B>], <C>]`. Or we won't support such cases?

lib/generator.js
```diff
@@ -1,38 +1,159 @@
...
+    );
+  }
+  return buf;
+};
+
+const generateRest = comments => {
+  const buf = [];
+  for (const comment of comments) {
+    if (comment.name === 'Example' || comment.name === 'Result') {
+      comment.comment = '\n```js\n' + comment.comment + '\n```';
```
lundibundi: Could we rename nested `comment` to smth like `source`? (i.e. `comment.source`) IIUC

lib/generator.js
```diff
@@ -1,38 +1,159 @@
...
+      comment.comment = ' ' + type + comment.comment;
+    }
+
+    comment.comment = comment.comment.split('\n')
+      .map(s => {
+        const trimmed = s.trimLeft();
+        return '\u00A0'.repeat(s.length - trimmed.length) + trimmed;
+      })
+      .join('  \n');
+    buf.push(`${comment.name}:${comment.comment}`);
```
lundibundi: Not sure but perhaps we need a space after a colon?

lib/introspection.js
```diff
@@ -1,91 +1,228 @@
...
+  const fnIndex = text.indexOf(fnData);
+  const lines = text.slice(0, fnIndex)
+    .split('\n')
+    .map(line => line.trim());
+  const comments = [];
+
+  for (let i = lines.length - 2; i >= 0; i--) {
+    const line = lines[i].trimLeft();
+    if (!line.startsWith('//') && line !== '//\n') break;
+    comments.unshift(line.slice(3));
```
lundibundi: Why `3`? (Perhaps this is the default indentation of a parameter, then could you put that in a constant)

test/example.js
```diff
@@ -1,22 +1,70 @@
...
-  const s = a2.join();
-  return a1 + s + a3;
-};
+/* eslint-disable no-unused-vars */
+
+// One-liner function description
+// Multi-line expanded function description. Note that all lines should have
+// length of 80 or less characters
+//   num <number> argument description
+//   str - <string>, dash before and comma after types are optional
```
lundibundi: Here you say that they are optional but at the bottom there is
> Current Metarhia style does not allow dash before and comma after types.

?

lib/generator.js
SemenchenkoVitaliy: I think we can support this kind of syntax.

lib/introspection.js
SemenchenkoVitaliy: 3 is two slashes and space and every comment line should start with `// `.

test/example.js
SemenchenkoVitaliy: They are indeed optional. I think that comments at the bottom should be removed from here and moved to the gist with our comments style because here it may sound confusing.

lib/generator.js
```diff
@@ -1,37 +1,176 @@
...
+];
+
+const types = [
+  ...primitives,
+  ...globalObjects,
+  'Primitive',
+];
+
+const baseUrl = 'https://developer.mozilla.org/en-US/docs/';
+const structuresBaseUrl = 'Web/JavaScript/Data_structures#';
```
lundibundi: Also, primitives..?

bin/cli.js
```diff
@@ -8,13 +8,28 @@ const common = require('metarhia-common');
 
 const metaschema = require('..');
 
-const cwd = process.cwd();
-const apiFile = path.resolve(cwd, process.argv[2]);
+let apiFile = path.resolve(process.cwd(), process.argv[2]);
+const packageFile = path.join(apiFile, 'package.json');
+
+if (fs.statSync(apiFile).isDirectory() && fs.existsSync(packageFile)) {
+  apiFile = path.join(apiFile, JSON.parse(fs.readFileSync(packageFile)).main);
```
belochub: I don't think there is any need to check if you were provided a directory with the `package.json` since you are using `require()` later which already supports loading a directory. Correct me if I'm wrong.

bin/cli.js
```diff
@@ -8,13 +8,28 @@ const common = require('metarhia-common');
...
 try {
   const imports = require(apiFile);
   const namespace = path.basename(apiFile, '.js');
-  const inventory = metaschema.introspect({ [namespace]: imports });
+  let data = '';
+  for (const module in require.cache) {
+    if (!module.includes('node_modules') &&
+      module !== __filename &&
+      fs.existsSync(module)) {
+      data += fs.readFileSync(module, 'utf8') + '\n';
```
belochub: It can be a binary file in case you are running it on a package with the native addons.
To avoid loading binary addons here, you can probably skip the files with `.node` extension.

lib/generator.js
```diff
@@ -1,37 +1,176 @@
...
+  'Date',
+  'Function',
+  'Map',
+  'WeakMap',
+  'Set',
+  'WeakSet',
+  'Int8Array',
+  'Uint8Array',
+  'Uint8ClampedArray',
+  'Float32Array',
```
belochub: This list seems to be also missing multiple other common global objects, such as `RegExp`, `ArrayBuffer`, `DataView`, `Int16Array`, `Uint16Array`, `Int32Array`, `Uint32Array`, `Float32Array`, `Float64Array`.

lib/generator.js
```diff
@@ -1,37 +1,176 @@
...
+  'RangeError',
+  'SyntaxError',
+  'ReferenceError',
+];
+
+const types = [
+  ...primitives,
+  ...globalObjects,
+  'Primitive',
+];
```
belochub: Why doesn't this list match the part of the list you pass to the `ALL_TYPES` `Set` constructor in the beginning of the `lib/introspection.js`?
Should this maybe be declared in one place only?

lib/introspection.js
```diff
@@ -1,91 +1,235 @@
...
+  'Deprecated:', // reasons for deprecation or previous names
+];
+
+// Search `text` for `fn` occurrences
+//   fn - <Function> to be searched
+//   text - <string> to be searched in
+// Returns: <Array> lines of comments before function
+const parseComments = (fn, text) => {
+  const fnData = fn.toString();
+  const fnIndex = text.indexOf(fnData);
```
belochub: You may not find the function if you try this on the package with the native addon, and this won't work correctly in that case.
Thus, you should handle not being able to find the function here.

lib/introspection.js
```diff
@@ -1,91 +1,235 @@
...
+    .split('\n')
+    .map(line => line.trim());
+  const comments = [];
+
+  for (let i = lines.length - 2; i >= 0; i--) {
+    const line = lines[i].trimLeft();
+    if (!line.startsWith('//')) break;
+    comments.unshift(line.slice(3));
+  }
+  return comments.map(line => line.trimRight()).filter(line => line.length);
```
belochub: I'd prefer an explicit check for the `length` not being equal to 0:
```suggestion
  return comments.map(line => line.trimRight())
    .filter(line => line.length !== 0);
```

lib/introspection.js
```diff
@@ -1,91 +1,235 @@
...
   }
-  return signature;
+
+  return sig;
 };
 
+// Parse function signature
+//   fn - <Function> method
+//   text - <string> data to parse
+// Returns: <Object> function signature
```
belochub: You should specify the fields in the returned object here.

test/example.md
```diff
@@ -1,20 +1,89 @@
...
 
-Returns: string, comment
+`methodName(num, str, arg, flag, arr, data, obj, cb)`
+  - `num: `[`<number>`][] -  argument description
+  - `str: `[`<string>`][] - dash before and comma after types are optional
+  - `arg: `[`<Array>`][]` | `[`<Date>`][]` | `[`<Object>`][] -  argument can have multiple types
+  - `flag: `[`<boolean>`][] -  description of argument can have multiple lines, but
+        there should be indentation of 4 spaces after beginning of
+        the first line
+  - `arr: <string[]>` -  square braces can be added after `type` to
```
belochub: This should also link to the same place as `<string>`.

test/example.js
```diff
@@ -1,22 +1,90 @@
...
+//   false, ['6'], data, {});
+//
+// Result:
+// {
+//   numArr: [4, 5, 6],
+//   strArr: ['4', '5', '6'],
+// }
+const methodName = (num, str, arg, flag, arr, data, obj, cb) => {};
+
+// List of standart types:
```
belochub:
```suggestion
// List of standard types:
```

test/example.js
```diff
@@ -1,22 +1,90 @@
...
+//         should increase by 2 more spaces for each level
+//       field2 <any>
+//     field3 <symbol> comment is optional, but type is  obligatory
+//   cb <Function> function description and/or example of
+//       usage. e.g. cb(arg1, ...arg2)
+//     arg1 <Map> arguments of function can also be nested using the
+//         same rules provided for `Object`
+//     arg2 <Array>
+// Returns: <Object> description of returned value. If this is
+//     an object or function with defined structure, than it should
```
belochub:
```suggestion
//     an object or function with defined structure, then it should
```

test/example.js
```diff
@@ -1,22 +1,90 @@
...
+//       the first line
+//   arr <string[]> square braces can be added after `type` to
+//       specify that this is array of `type`s
+//   data <CustomType> custom type can be specified. e.g. this can be
+//       instance of some class
+//   obj <Object> its description
+//     field1 <Object> fields of object can be nested but indentation
+//         should increase by 2 more spaces for each level
+//       field2 <any>
+//     field3 <symbol> comment is optional, but type is  obligatory
```
belochub:
```suggestion
//     field3 <symbol> comment is optional, but type is obligatory
```

test/example.md
```diff
@@ -1,20 +1,100 @@
...
 
-Returns: string, comment
+Multi-line expanded function description. Note that all lines should have
+length of 80 or less characters
 
-Hint: message
+`methodName(num, str, arg, flag, arr, data, obj, cb)`
+  - `num: `[`<number>`] - argument description
+  - `str: `[`<string>`] - dash before and comma after types are optional
+  - `arg: `[`<Array>`]` | `[`<Date>`]` | `[`<Object>`] - argument can have multiple types
```
nechaido: I don't like the fact that this line is 89 characters long.

SemenchenkoVitaliy: Previously, all indentation was preserved but @lundibundi suggested removing it so that markdown will manage everything. IMO, all formatting in code comments should stay in docs.

Issue #110, opened 2018-09-21
Rewrite tests


bugagashenkj:
Rewrite tests https://github.com/metarhia/metaschema/tree/master/test using metatests(https://github.com/metarhia/metatests).

PR #111, opened 2018-09-24
Implement new tests

bugagashenkj:
#110


test/database.js
```diff
@@ -1,17 +1,80 @@
 'use strict';
 
+/* eslint new-cap: 0 */
+const metatests = require('metatests');
 const metaschema = require('..');
+const decorators = require('../lib/decorators');
```
tshemsedinov: `{ Include, ... } = require('../lib/decorators').attribute;`

test/database.js
```diff
@@ -1,17 +1,80 @@
...
+  metaschema.load('general', (err, schema) => {
+    test.assertNot(err);
+    metaschema.build(schema);
+
+    for (const categoryName in categoriesConfigs) {
+      const config = categoriesConfigs[categoryName];
+      const category = metaschema.categories.get(categoryName);
+
+      test.strictSame(config.name, category.name);
+      for (const key in config.definition) {
```
lundibundi: Wouldn't `test.strictSame(category.definition, config.definition)` be enough?

test/database.js
```diff
@@ -1,17 +1,80 @@
...
+metatests.test('Database / database', (test) => {
+  metaschema.load('general', (err, schema) => {
+    test.assertNot(err);
+    metaschema.build(schema);
+
+    for (const categoryName in categoriesConfigs) {
+      const config = categoriesConfigs[categoryName];
+      const category = metaschema.categories.get(categoryName);
+
+      test.strictSame(config.name, category.name);
```
lundibundi: Argument order is incorrect, its `actual, expected`. This should be `test.strictSame(category.name, config.name);`

test/database.js
```diff
@@ -1,17 +1,80 @@
...
+    Languages: Many({
+      category: 'Language',
+      definition: categoriesConfigs.Language.definition
+    })
+  }
+};
+
+metatests.test('Database / database', (test) => {
+  metaschema.load('general', (err, schema) => {
+    test.assertNot(err);
```
lundibundi: test.error(err);

test/database.js
```diff
@@ -1,17 +1,80 @@
...
+    Sex: domains.sex,
+    Born: domains.born,
+    Languages: Many({
+      category: 'Language',
+      definition: categoriesConfigs.Language.definition
+    })
+  }
+};
+
+metatests.test('Database / database', (test) => {
```
lundibundi: Test case name doesn't seem correct.

test/guard.js
```diff
@@ -15,9 +16,9 @@ const buyTickets = (
   return true;
 };
 
-buyTickets(101, 2, 'Kiev, Pobedy 37');
-try {
-  buyTickets(101, '2', 37);
-} catch (e) {
-  console.log(e.message);
-}
+metatests.case('Metaschema / guard', { buyTickets }, {
```
belochub: This doesn't really test anything since `guard` is always returning `true`.

test/structure.js
```diff
@@ -1,61 +1,56 @@
 'use strict';
 
+const metatests = require('metatests');
 const metaschema = require('..');
 
 metaschema.load('geometry', (err, schema) => {
   if (err) throw err;
```
belochub: This should be tested as well.

lundibundi:
@bugagashenkj could you rebase this and update to use metatests@0.2.1?

test/structure.js
```diff
@@ -1,61 +1,59 @@
 'use strict';
 
+const metatests = require('metatests');
 const metaschema = require('..');
 
 metaschema.load('geometry', (err, schema) => {
```
lundibundi: `metatests` don't support such usage and this may lead to multiple 'results' printed in the end.
All tests should be added synchronously. Therefore this must be changed to smth like
```js
const geometryTest = metatests.test('geometry');
metaschema.load('geometry', (err, schema) => {
  // ... use geometryTest and/or metaschema.case ...
  // ...
  geometryTest.end();
})
```

Or
```js
metatests.test('geometry', (test) => {
  metaschema.load('geometry', (err, schema) => {
    // ... use test and/or metaschema.case ...
    // ...
    geometryTest.end();
  })
});
```

Whichever you prefer.

test/structure.js
```diff
@@ -1,61 +1,59 @@
...
-  const pl1 = geometry('Polyline', {
-    points: [
-      { x: 1, y: 3 },
-      { x: 2, y: 2 },
-      { x: 3, y: 4 },
-      { x: 4, y: 5 },
-      { x: 3, y: 5 }
+  const geometry = metaschema.build(schema);
+  metatests.case('Metaschema / structure', { metaschema, geometry }, {
+    'metaschema.createInstance': [
```
lundibundi: Could probably be inlined to `createInstance` to be brief. (with appropriate change to the namespace argument)

test/structure.js
```diff
@@ -1,61 +1,59 @@
...
+      ['Point', { x: 3, y: 5 }, { x: 3, y: 5 }]
+    ],
+    'geometry': [
+      ['Point', { x: 3, y: 5 },        { x: 3, y: 5 }],
+      ['Point', [3, 5],                { x: 3, y: 5 }],
+      ['Point', 3, 5,                  { x: 3, y: 5 }],
+      ['Point', { x: 'hello', y: 2 },  null],
+      ['Point', { name: 'Marcus' },    {}],
```
lundibundi: Shouldn't this be `null`?
Looks like a bug as those fields are `required`.

test/validate.js
```diff
@@ -1,9 +1,12 @@
 'use strict';
 
+const metatests = require('metatests');
 const metaschema = require('..');
 
 const path = './schemas/metaschema/StructureField.schema';
 
+const getDefinition = name => metaschema.categories.get(name).definition;
+
 metaschema.loadSchema(path, (err, schema) => {
```
lundibundi: Same as in `test/structure.js`.

test/validate.js
```diff
@@ -14,46 +17,33 @@ metaschema.loadSchema(path, (err, schema) => {
...
-  // Validate category field schema
-  {
-    const { definition } = metaschema.categories.get('StructureField');
-    const valid = metaschema.validateFields('StructureField', definition);
-    console.dir(valid);
-  }
-
+  metatests.case('Metaschema / validate', { metaschema }, {
+    'metaschema.validate': [
+      [
```
lundibundi: Could you also add
`'Schema1', {}, { valid: true, errors: [] }`
and
`'Schema2', {}, { valid: false, errors .... }`
?

lundibundi:
ping @bugagashenkj

test/structure.js
bugagashenkj: Should I do something about this?

bugagashenkj:
@lundibundi please approve.

package.json
```diff
@@ -23,7 +23,7 @@
   },
   "readmeFilename": "README.md",
   "scripts": {
-    "test": "npm run lint && npm run docs && node ./test/all.js",
+    "test": "eslint . && metatests test/",
```
lundibundi: Why did you remove the previous changes? This should be
`"test": "npm run lint && npm run docs && metatests test/"`.

test/structure.js
lundibundi: You can go ahead and change this as that PR is going to be accepted, there is only a minor change waiting.

lundibundi:
Landed in https://github.com/metarhia/metaschema/pull/141 (https://github.com/metarhia/metaschema/pull/141/commits/bbe16021cd339eba7e33e189d7a07047c5767b30).

PR #114, opened 2018-09-28
Validate schemas on loading

### LINKED ISSUE #112: Deny relation from Global to Local
Deny relation from Global to Local.
`Registry, Dictionary, System, Log, History` decorated schemas must not have relations to `Local`

nechaido:
Closes #112.

### DIFF
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -18,6 +18,35 @@ const categories = new Map();
 
 const structs = {};
 
+const ValidationErrorSerializer = {
+  linkToLog: (error) =>
+    'Forbidden reference to a \'Log\' category ' +
+    `from ${error.source}.${error.property}`,
+  illegalLinkToLocal: (error) =>
+    `Illegal reference to a 'Local' category '${error.info.destination}' ` +
+    `from ${error.source}.${error.property}`,
+  unresolvedDomain: (error) =>
+    `Reference to an unresolved domain '${error.info.domain}' ` +
+    `from ${error.source}.${error.property}`,
+  unresolvedCategory: (error) =>
+    `Reference to an unresolved category '${error.info.category}' ` +
+    `from ${error.source}.${error.property}`,
+};
+
+class ValidationError extends Error {
+  constructor(type, source, property, info) {
+    super();
+    this.type = type;
+    this.source = source;
+    this.property = property;
+    this.info = info;
+  }
+
+  toString() {
+    return ValidationErrorSerializer[this.type](this);
+  }
+}
+
 const processSchemaFile = (fileName, source) => {
   const sandbox = {};
   Object.assign(sandbox, structs);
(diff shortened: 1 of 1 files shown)


lib/schema.js
```diff
@@ -77,23 +77,73 @@ const processSchema = (
   }
 };
 
+const getCategoryType = (category) => {
+  const type = category.constructor.name;
+  return type === 'Object' ? 'Local' : type;
```
tshemsedinov: Do you mean this will return `Local` for all non-decorated categories? Need to be commented here and documented later.

lib/schema.js
```diff
@@ -77,23 +77,73 @@ const processSchema = (
...
+  if (destinationType === 'Log') {
+    return `Illegal reference to Log from ${sourceName}.${propertyName}`;
+  }
+
+  if (destinationType === 'Local' && sourceType !== 'Local') {
+    return `Illegal reference to Local category ${destination} from ` +
+      `${propertyName} ${sourceName}.${propertyName}`;
+  }
+
+  return null;
```
tshemsedinov: `return undefined;` or simply remove this line

lib/schema.js
```diff
@@ -77,23 +77,73 @@ const processSchema = (
...
-        if (category) field.definition = category;
+        if (category) {
+          const error = verifyLink(
+            schema, schemaName, category, field.category, fieldName
+          );
+          if (!error) {
+            field.definition = category;
+          } else {
+            errors.push(error);
+          }
```
tshemsedinov: in such cases I'd prefer:
```js
if (error) errors.push(error);
else field.definition = category;
```

lib/schema.js
nechaido: On the other hand, should we, maybe, wrap every undecorated category into `Local` on schema loading?

lib/schema.js
nechaido: I don't like that style at all, moreover this way `else` has the same indentation as `else` one level higher, which is a bit confusing.
```js
if (category) {
  // ...
  if (!error) field.definition = category;
  else errors.push(error);
} else {
  // ...
}

```

lib/schema.js
nechaido: This function now returns an object, so `null` is better.

lib/schema.js
```diff
@@ -104,20 +133,16 @@ const verifyLink = (
...
   }
 
   if (destinationType === 'Local' && sourceType !== 'Local') {
-    return {
-      type: 'illegalLinkToLocal',
-      source: sourceName,
-      destination: destinationName,
-      property: propertyName,
-    };
+    return new Error(
```
belochub: This must probably be `new ValidationError`.

lib/schema.js
```diff
@@ -18,6 +18,35 @@ const categories = new Map();
 
 const structs = {};
 
+const ValidationErrorSerializer = {
+  linkToLog: (error) =>
+    'Forbidden Reference to a \'Log\' category ' +
```
belochub: Why is 'Reference' capitalized here?

lundibundi:
@belochub ping.

belochub:
Landed in https://github.com/metarhia/metaschema/commit/734a63378906f7243a91addb4fc72dbd3ff5bb7c.

PR #121, opened 2018-10-11
Fix validate not returning the result

belochub:


belochub:
Landed in https://github.com/metarhia/metaschema/commit/b5bd6a0d6bc85421adb336d44abbd1a43bb8b3b0.

PR #124, opened 2018-10-16
Add domains for ISO country codes

tshemsedinov:


schemas/metaschema/domains.schema
```diff
@@ -28,6 +28,9 @@
   HashSet: { type: 'object', class: 'HashSet' },
   JSON: { type: 'string', control: 'Multiline' },
   HTML: { type: 'string', control: 'Multiline' },
+  CountryAlpha2: { type: 'string', length: 2, comment: 'ISO 3166-1 alpha-2' },
+  CountryAlpha3: { type: 'string', length: 3, comment: 'ISO 3166-1 alpha-3' },
+  CountryNumeric: { type: 'string', comment: 'ISO 3166-1 numeric' },
```
belochub: This should also have length 3 since ISO 3166-1 numeric codes are three-digit country codes.

Issue #118, opened 2018-10-11
Can not create instance of domains which are value decorators  


o-rumiantsev:
Function `createInstance` in `lib/schema.js` can not create instances of domains which are value decorators, because we can not specify `type` field for them. 

https://github.com/metarhia/metaschema/blob/aac7ad16ce34cc6a066f850f5a99ddcb9bcf232f/lib/schema.js#L137-L141

Here an identifier `type` is an own JS type of `value`, but `expected` is `name`(name of the category, we creating instance of), because value decorators `Enum` and `Flags` can not have property `type`. So `type === expected` is `false` and it returns `null`

tshemsedinov:
Yes, please fix this adding additional check if `domain.constructor.name` will be `Enum` or `Flags` we need to validate `value` to be one of allowed in `Enum` and `Flags` respectively. @o-rumiantsev

PR #125, opened 2018-10-17
Rewrite createInstance

o-rumiantsev:
Closes: https://github.com/metarhia/metaschema/issues/118

### DIFF
--- a/lib/schema-processors.js
+++ b/lib/schema-processors.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const { Enum, iter } = require('@metarhia/common');
+const { Enum, Flags, iter } = require('@metarhia/common');
 
 const {
   getCategoryType,
@@ -10,6 +10,8 @@ const {
 } = require('./schema-utils');
 
 const { SchemaValidationError, MetaschemaError } = require('./schema-errors');
+
+const domainConstructors = { Enum, Flags };
 
 // Verifies that there could be link from source category to destination
 //   source - <Object>
@@ -37,6 +39,19 @@ const verifyLink = (source, destination, propertyName) => {
   }
 
   return null;
+};
+
+// Add a Class property to a definition of Enum and Flags
+//   ms - <Metaschema>
+const processDomains = ms => {
+  for (const definition of ms.domains.values()) {
+    const domainKind = extractDecorator(definition);
+    const domainConstructor = domainConstructors[domainKind];
+    if (domainConstructor) {
+      const values = definition.values;
+      definition.Class = domainConstructor.from(...values);
+    }
+  }
 };
 
 // Crosslink loaded categories
@@ -320,6 +335,7 @@ const processDisplayModes = ms => {
 };
 
 module.exports = {
+  processDomains,
   processCategories,
   processForms,
   processViews,
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const { iter } = require('@metarhia/common');
+const { iter, Int64, Uint64, Money } = require('@metarhia/common');
 
 const {
   SchemaValidationError,
@@ -11,6 +11,7 @@ const {
 const { REFERENCE_TYPES, extractDecorator } = require('./schema-utils');
 
 const {
+  processDomains,
   processCategories,
   processActions,
   processViews,
@@ -20,6 +21,41 @@ const {
 
 const factorify = Symbol('factorify');
 const validate = Symbol('validate');
+
+const classWrappers = {
+  Money: value => {
+    try {
+      return new Money(value);
+    } catch (e) {
+      return null;
+    }
+  },
+  Uint8Array: value => {
+    try {
+      return Uint8Array.from(value);
+    } catch (e) {
+      return null;
+    }
+  },
+  HashMap: value => {
+    try {
+      return new Map(value);
+    } catch (e) {
+      return null;
+    }
+  },
+  HashSet: value => {
+    try {
+      return new Set(value);
+    } catch (e) {
+      return null;
+    }
+  },
+  Date: value => {
+    const date = new Date(value);
+    return Number.isNaN(Number(date)) ? null : date;
+  },
+};
 
 const domainValidators = {
   string: (domain, prop, value) => {
@@ -76,9 +112,9 @@ const domainValidators = {
 };
 
 // Validates value against a domain
-//   value - <any>
-//   path - <string>
-//   domain - <Object>
+//   value <any>
+//   path <string>
+//   domain <Object>
 // Returns: <ValidationError[]>
 const validateDomain = (value, path, domain) => {
   const errors = [];
@@ -186,41 +222,52 @@ class Metaschema {
     this.sources = [];
   }
 
-  // Create category instance
-  //   def - <Object> | <string> field definition
-  //   value - <any> value to check/create
-  // Returns: instance of `def`
-  createInstance(def, value) {
-    let name;
-    if (typeof def === 'string') name = def;
-    else name = def.category || def.domain;
-
-    const decorator = extractDecorator(def);
-    const category = this.categories.get(name);
-    if (category && category.factory) {
-      return category.factory(value);
-    }
-    if (decorator === 'List') {
-      const cat = this.categories.get(def.category);
-      if (!cat) return null;
-      const checked = value.map(v => this.createInstance(cat, v));
-      if (checked.some(v => v === null)) return null;
-      return checked;
-    }
+  // Create domain instance
+  //   name <string> domain name
+  //   value <any> value to check/create
+  // Returns: instance of domain
+  createDomainInstance(name, value) {
     const domain = this.domains.get(name);
+    const decorator = extractDecorator(domain);
+    if (domain.normalize) value = domain.normalize(value);
+    if (domain.parse) return domain.parse(value);
+    if (decorator === 'Enum') {
+      const valid = domain.Class.has(value);
+      return valid ? domain.Class.from(value) : null;
+    }
+    if (decorator === 'Flags') {
+      if (typeof value === 'string' || typeof value === 'number') {
+        return domain.Class.from(new Uint64(value));
+      }
+      const valid =
+        Array.isArray(value) && value.every(v => domain.Class.has(v));
+      return valid ? domain.Class.from(...value) : null;
+    }
     const type = typeof value;
-    const expected = domain ? domain.type : name;
-    if (type === expected) return value;
-    return null;
+    const expected = domain.type;
+    if (expected === 'bigint') return new Int64(value);
+    if (expected === 'object') {
+      const classWrapper = classWrappers[domain.class];
+      return classWrapper ? classWrapper(value) : null;
+    }
+    return type === expected ? value : null;
+  }
+
+  // Create category instance
+  //   name <string> category name
+  //   value <any[]> value to check/create
+  // Returns: instance of category
+  createInstance(name, ...args) {
+    if (!this.categories.has(name)) return null;
+    return this.categories.get(name).factory(...args);
   }
 
   // Create factory from category definition
-  //   name - <string> category name
-  //   definition - <Object> category definition
+  //   name <string> category name
+  //   definition <Object> category definition
   // Returns: function, (...args)
   [factorify](name, definition) {
     const properties = Object.keys(definition);
-    const required = new Set(properties.filter(p => definition[p].required));
     const factory = (...args) => {
       let instance = args[0];
       if (args.length > 1) instance = args;
@@ -248,14 +295,28 @@ class Metaschema {
           value = def(value);
           obj[field] = value;
         } else {
-          value = this.createInstance(def, value);
+          if (def.category) {
+            if (typeof value === 'object') {
+              value = this.createInstance(def.category, value);
+            } else {
+              value = new Int64(value);
+            }
+          }
+          if (def.domain) value = this.createDomainInstance(def.domain, value);
           if (value !== null) obj[field] = value;
           else return null;
         }
       }
-      for (const prop of required) {
-        if (obj[prop] === null || obj[prop] === undefined) {
-          return null;
+      for (const prop of properties) {
+        const def = definition[prop];
+        if (!obj.hasOwnProperty(prop)) {
+          if (def.default) {
+            const value = this.createDomainInstance(def.domain, def.default);
+            if (value === null) return null;
+            obj[prop] = value;
+          } else if (def.required) {
+            return null;
+          }
         }
       }
       return obj;
@@ -264,11 +325,11 @@ class Metaschema {
   }
 
   // Validate object against a schema
-  //   schema - <Object>, schema to validate against
-  //   object - <Object>, object to validate
-  //   patch - <boolean>, flag to determine if the object contains patch or
+  //   schema <Object>, schema to validate against
+  //   object <Object>, object to validate
+  //   patch <boolean>, flag to determine if the object contains patch or
   //       value, default: `false`
-  //   path - <string>, path to an object, for nested objects, default: `''`
+  //   path <string>, path to an object, for nested objects, default: `''`
   // Returns: <ValidationError[]>
   [validate](schema, object, patch = false, path = '') {
     const errors = [];
@@ -333,11 +394,11 @@ class Metaschema {
   }
 
   // Validate instance against a category
-  //   categoryName - <string>, category name
-  //   instance - <any>, instance to validate
-  //   patch - <boolean>, flag to determine if the object contains patch or
+  //   categoryName <string>, category name
+  //   instance <any>, instance to validate
+  //   patch <boolean>, flag to determine if the object contains patch or
   //       value, default: `false`
-  //   path - <string>, path to an object, for nested objects, default: `''`
+  //   path <string>, path to an object, for nested objects, default: `''`
   // Returns: <MetaschemaError> | <null>
   validateCategory(categoryName, instance, patch = false, path = '') {
     const category = this.categories.get(categoryName);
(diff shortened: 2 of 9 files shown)


lib/schema.js
```diff
@@ -206,7 +216,12 @@ const createInstance = (
   const domain = domains.get(name);
   const type = typeof value;
   const expected = domain ? domain.type : name;
-  if (type === expected) return value;
+  const domainType = domain ? domain.constructor.name : null;
+  const isValueDecorator = domainType in valueDecoratorValidators;
+  const validate = valueDecoratorValidators[domainType];
+  if (type === expected || isValueDecorator && validate(domain, value)) {
+    return value;
```
tshemsedinov: Double check this returns scalar value, because `createInstance` intended to duplicate category calling factory and pass  copies of scalar values because scalars are passing by value. Do we pass flags and enums by value?

o-rumiantsev: Not at all, we pass flags as array of values

tshemsedinov: array will be passed by reference

o-rumiantsev:
Need https://github.com/metarhia/common/pull/201 to be merged to use `common.Flags`

lib/schema.js
```diff
@@ -151,9 +152,28 @@ class Metaschema {
...
+      const { values } = def;
+      const valid = Enum.from(values).has(value);
+      if (!valid) return null;
+      return value;
+    }
+    if (decorator === 'Flags') {
+      const { values } = def;
+      let flags;
+      try {
+        flags = Flags.from(...values).from(...value);
```
lundibundi: @o-rumiantsev I think it will be better to use `has`
```suggestion
        const flags = Flags.from(...values);
        if (value.some(v => !flags.has(v)) return null;
```
Though this will only work for array so perhaps you should use `iter`.

o-rumiantsev: `Flags.from`  returns a `FlagsClass` class, so in your suggestion variable `flags` wouldn't have method `has`

lundibundi: I thought `has` is static, if it isn't I think we should make it so.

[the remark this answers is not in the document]
nechaido: I'd prefer to store it under `Class` property so that it can be used like this:
```js
const value = new this.domains.get(domainName).Class(...value);
```

lundibundi:
> IMHO, createInstance should normalize and parse all values (eq. all links should > be casted into Int64. enum to corresponding class and so on), save them into an > object, than call validate on a created object, to ensure all values are valid.
>
> Also I'd prefer to spit domain and category creation into separate functions.

+1.

lib/schema.js
```diff
@@ -13,6 +12,29 @@ const addSchema = Symbol('addSchema');
...
 const validate = Symbol('validate');
+const classifyDomains = Symbol('classifyDomains');
+
+const domainConstructors = { Enum, Flags };
+const classWrappers = {
+  Uint8Array: value => Uint8Array.from(value),
+  HashMap: value => {
+    try {
+      const map = new Map(value);
+      return map;
```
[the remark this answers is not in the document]
nechaido: I think that this should be `null`, not an empty `Map` because value cannot be converted to a Map.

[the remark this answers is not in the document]
nechaido: IFAIU, `try {} catch` is in case someone attempts to pass a value that cannot be converted to `Map` in that case constructor will throw but we need to return `null`, no raise an Error.

test/schema.js
```diff
@@ -50,4 +56,170 @@ metaschema.fs.load(null, null, (err, schemas) => {
       const actual = ms.createInstance('schema', {});
       test.strictSame(actual, null);
     });
+
+  const dupSchemas = duplicate(schemas);
+  const [domains] = dupSchemas;
+
+  domains[1].BigInt = { type: 'bigint' };
```
nechaido: Isn't it better to put schemas into a fixtures directory, and load them the way they are supposed to load?

lib/schema.js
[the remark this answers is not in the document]
nechaido: In case your point is: "You can combine this lith with the previous one into: `return new Map(value)`" - you have +1 for that.

lib/schema.js
```diff
@@ -311,21 +333,43 @@ class Metaschema {
...
       if (checked.some(v => v === null)) return null;
       return checked;
     }
+    if (decorator === 'Enum') {
+      const valid = def.Class.has(value);
+      return valid ? def.Class.from(value) : null;
+    }
+    if (decorator === 'Flags') {
+      const valid =
+        Array.isArray(value) &&
```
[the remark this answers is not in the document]
nechaido: @lundibundi this should be resolved now.

lib/schema.js
```diff
@@ -302,34 +325,57 @@ class Metaschema {
...
-    const category = this.categories.get(name);
-    if (category && category.factory) {
-      return category.factory(value);
-    }
+  createDomainInstance(def, value) {
+    if (def.domain === 'Id') return new Uint64(value);
+    if (def.domain === 'Money') return new Money(value);
+    const domain = this.domains.get(def.domain);
+    const decorator = extractDecorator(domain);
+    if (domain.normalize) value = domain.normalize(value, null, domain);
```
nechaido:
```suggestion
    if (domain.normalize) value = domain.normalize(value);
```

lib/schema.js
```diff
@@ -302,34 +325,57 @@ class Metaschema {
...
-    else name = def.category || def.domain;
-
-    const decorator = extractDecorator(def);
-    const category = this.categories.get(name);
-    if (category && category.factory) {
-      return category.factory(value);
-    }
+  createDomainInstance(def, value) {
+    if (def.domain === 'Id') return new Uint64(value);
+    if (def.domain === 'Money') return new Money(value);
```
nechaido: `Money` should be handled by `classWrappers`.

lib/schema.js
```diff
@@ -302,34 +325,57 @@ class Metaschema {
...
-    if (typeof def === 'string') name = def;
-    else name = def.category || def.domain;
-
-    const decorator = extractDecorator(def);
-    const category = this.categories.get(name);
-    if (category && category.factory) {
-      return category.factory(value);
-    }
+  createDomainInstance(def, value) {
+    if (def.domain === 'Id') return new Uint64(value);
```
nechaido: This should be handled by `Id`'s `parse` property.

lib/schema.js
```diff
@@ -302,34 +325,57 @@ class Metaschema {
...
+    if (def.domain === 'Money') return new Money(value);
+    const domain = this.domains.get(def.domain);
+    const decorator = extractDecorator(domain);
+    if (domain.normalize) value = domain.normalize(value, null, domain);
+    if (domain.parse) value = domain.parse(value);
     if (decorator === 'List') {
-      const cat = this.categories.get(def.category);
+      const cat = this.categories.get(domain.category);
```
nechaido: `domain.category` looks weird, you should probably return `null` in this case.

lib/schema.js
```diff
@@ -302,34 +325,57 @@ class Metaschema {
...
+    }
     const type = typeof value;
-    const expected = domain ? domain.type : name;
+    const expected = domain.type;
+    if (expected === 'bigint') return new Int64(value);
+    if (expected === 'object') {
+      const classWrapper = classWrappers[domain.class];
+      return classWrapper ? classWrapper(value) : null;
+    }
     if (type === expected) return value;
```
nechaido:
```suggestion
    return type === expected ? value : null;
```

lib/schema.js
```diff
@@ -302,34 +325,57 @@ class Metaschema {
...
     return null;
   }
 
+  // Create category instance
+  //   def - <Object> field definition
```
nechaido: This should be just a string.

lib/schema.js
```diff
@@ -302,34 +325,57 @@ class Metaschema {
     }
   }
 
-  // Create category instance
-  //   def - <Object> | <string> field definition
+  // Create domain instance
+  //   def - <Object> field definition
```
nechaido: This should be just a string, domain name.

test/database.js
```diff
@@ -4,6 +4,7 @@
 
 const path = require('path');
 const metatests = require('metatests');
+const { Enum: CommonEnum } = require('@metarhia/common');
```
nechaido:
```suggestion
const common = require('@metarhia/common');
```

test/database.js
```diff
@@ -29,6 +30,8 @@ const domains = {
   },
 };
 
+domains.sex.definition.Class = CommonEnum.from('female', 'male');
```
nechaido:
```suggestion
domains.sex.definition.Class = common.Enum.from('female', 'male');
```

lib/schema.js
o-rumiantsev: In this case there is no need in `createInstance` for category, because we have `buildCategory` which does absolutely the same

lib/schema.js
o-rumiantsev: Maybe it would be better to move it to `createInstance`? I guess `List` might be used like this:
```javascript
{
  field: List('SomeCategory') 
  // in definition it will look like List { category: 'SomeCategory' }
}
```

lib/schema.js
nechaido: We can delete `buildCategory` in that case.

lib/schema.js
```diff
@@ -383,7 +420,9 @@ class Metaschema {
           value = def(value);
           obj[field] = value;
         } else {
-          value = this.createInstance(def, value);
+          const name = def.category || def.domain;
+          if (def.category) value = this.createInstance(name, value);
+          if (def.domain) value = this.createDomainInstance(name, value);
```
lundibundi: Why not
```js
          if (def.category) value = this.createInstance(def.category, value);
          if (def.domain) value = this.createDomainInstance(def.domain, value);
```
?

o-rumiantsev: It seems prettier for me

lib/schema.js
```diff
@@ -383,21 +418,40 @@ class Metaschema {
...
           if (value !== null) obj[field] = value;
           else return null;
         }
       }
-      for (const prop of required) {
-        if (obj[prop] === null || obj[prop] === undefined) {
-          return null;
+      for (const prop of properties) {
+        const def = definition[prop];
+        if (def.default && !(prop in obj)) {
```
nechaido:
```js
if (obj.hasOwnProperty(prop)) {
  if (def.default) {
    // ...
  } else if (def.required) {
    return null;
  }
}
```

lib/schema.js
```diff
@@ -428,21 +464,41 @@ class Metaschema {
...
       }
-      for (const prop of required) {
-        if (obj[prop] === null || obj[prop] === undefined) {
-          return null;
+      for (const prop of properties) {
+        const def = definition[prop];
+        if (!obj.hasOwnProperty(prop)) {
+          if (def.default) {
+            let value = null;
+            value = this.createDomainInstance(def.domain, def.default);
```
```suggestion
            let value = this.createDomainInstance(def.domain, def.default);
```
?

lib/schema.js
```diff
@@ -428,21 +464,41 @@ class Metaschema {
...
           if (value !== null) obj[field] = value;
           else return null;
         }
       }
-      for (const prop of required) {
-        if (obj[prop] === null || obj[prop] === undefined) {
-          return null;
+      for (const prop of properties) {
+        const def = definition[prop];
+        if (!obj.hasOwnProperty(prop)) {
```
lundibundi: If `obj = { A: undefined }` but `A` is `required` this won't catch that and just ignore this property, is that ok?

lib/schema.js
```diff
@@ -428,21 +464,41 @@ class Metaschema {
...
+          if (def.default) {
+            let value = null;
+            value = this.createDomainInstance(def.domain, def.default);
+            if (value === null) return null;
+            obj[prop] = value;
+          } else if (def.required) {
+            return null;
+          }
```
lundibundi: Also, I'd merge those branches:
```js
if (!obj.hasOwnProperty(prop)) {
  if (def.default) {
    obj[prop] = this.createDomainInstance(def.domain, def.default);
  }
  if (obj[prop] == null && def.required) return null;
}
```

lib/schema.js
```diff
@@ -428,21 +464,41 @@ class Metaschema {
...
     };
     return factory;
   }
 
+  // Add a Class property to a definition of Enum and Flags
+  [classifyDomains]() {
+    for (const definition of this.domains.values()) {
+      const domainKind = extractDecorator(definition);
+      const domainConstructor = domainConstructors[domainKind];
+      if (!domainConstructor) continue;
```
lundibundi: IMO, it'd be better to just use `if` without continue, the branch will only have 2 lines.

lib/schema.js
```diff
@@ -13,6 +13,31 @@ const addSchema = Symbol('addSchema');
 const factorify = Symbol('factorify');
 const initActions = Symbol('initActions');
 const validate = Symbol('validate');
+const classifyDomains = Symbol('classifyDomains');
+
+const domainConstructors = { Enum, Flags };
+const classWrappers = {
+  Money: value => new Money(value),
+  Uint8Array: value => Uint8Array.from(value),
```
lundibundi: Why is there no try-catch block for the `Uint8Array` and `Money`?
i.e. `Uint8Array.from(null)` will throw TypeError.

lib/schema.js
```diff
@@ -367,32 +392,44 @@ class Metaschema {
     }
   }
 
-  // Create category instance
-  //   def - <Object> | <string> field definition
+  // Create domain instance
+  //   name - <string> domain name
```
tshemsedinov:
```suggestion
  //   name <string> domain name
```

tshemsedinov: For all comments

lib/schema.js
o-rumiantsev: We should return `null` even if `obj[prop] === null` and property is not required.

schemas/metaschema/domains.schema
```diff
@@ -1,5 +1,5 @@
 {
-  Id: { type: 'bigint', parse: x => new api.common.Uint64(x) },
+  Id: { type: 'bigint', parse: x => new common.Uint64(x) },
```
lundibundi: I think this should be reverted	as AFAIK we have decided to bundle any libs under `api` in schemas.
/cc @tshemsedinov @belochub @nechaido

lundibundi:
ping @o-rumiantsev.

lib/schema.js
```diff
@@ -175,41 +212,51 @@ class Metaschema {
...
+    }
+    return type === expected ? value : null;
+  }
+
+  // Create category instance
+  //   name <string> category name
+  //   value <any[]> value to check/create
+  // Returns: instance of category
+  createInstance(name, ...args) {
+    return this.categories.get(name).factory(...args);
```
nechaido: You should check that such category exists.

lib/schema.js
```diff
@@ -237,27 +284,47 @@ class Metaschema {
           value = def(value);
           obj[field] = value;
         } else {
-          value = this.createInstance(def, value);
+          if (def.category) value = this.createInstance(def.category, value);
```
nechaido: It should be possible to save an objects `id` instead of the whole object IMHO.

o-rumiantsev: So, from this schema:
```javascript
{
  field: { category: 'CategoryName' }
}
```
We can create such instance:
```javascript
{ 
  field: 1, // an id
}
```
?

nechaido: It should be possible to create: `{ field: new Id(1) }` from: `{ field: 1 }`.

lib/schema.js
```diff
@@ -19,6 +19,43 @@ const {
 
 const factorify = Symbol('factorify');
 const validate = Symbol('validate');
+const classifyDomains = Symbol('classifyDomains');
```
nechaido: This should be moved to `schema-processors` and renamed to `processDomains` to be consistent.

lib/schema.js
o-rumiantsev: What is an `Id` here. Is it included in `@metarhia/common` or I should define it by myself?

nechaido: You may use `Int64` for now.

lib/schema-processors.js
```diff
@@ -39,6 +41,19 @@ const verifyLink = (source, destination, propertyName) => {
   return null;
 };
 
+// Add a Class property to a definition of Enum and Flags
+//   ms - <Metaschema>
+const processDomains = ms => {
+  for (const definition of ms.domains.values()) {
+    const domainKind = extractDecorator(definition);
```
nechaido:
```suggestion
    const type = extractDecorator(definition);
```

lib/schema.js
```diff
@@ -186,41 +222,52 @@ class Metaschema {
...
+    }
+    return type === expected ? value : null;
+  }
+
+  // Create category instance
+  //   name <string> category name
+  //   value <any[]> value to check/create
+  // Returns: instance of category
+  createInstance(name, ...args) {
+    if (!this.categories.has(name)) return null;
```
nechaido:
```suggestion
    const category = this.categories.get(name));
    return category ? category.factoryfy(...args) : null;
```

lib/schema.js
```diff
@@ -248,14 +295,28 @@ class Metaschema {
...
           obj[field] = value;
         } else {
-          value = this.createInstance(def, value);
+          if (def.category) {
+            if (typeof value === 'object') {
+              value = this.createInstance(def.category, value);
+            } else {
+              value = new Int64(value);
+            }
+          }
```
nechaido:
```suggestion
          } else
```

lib/schema.js
```diff
@@ -248,14 +295,28 @@ class Metaschema {
           value = def(value);
           obj[field] = value;
         } else {
-          value = this.createInstance(def, value);
+          if (def.category) {
+            if (typeof value === 'object') {
```
nechaido: Ternary?

lib/schema.js
nechaido: IMHO, it is not, we should prohibit `null` or `undefined`.

o-rumiantsev:
This PR became heavily outdated, so I am closing it

Issue #130, opened 2018-10-20
Bug in reference validation


tshemsedinov:
Please test validation on globalstorage. For example `Action.Category` references `Category`, but it is not `Local`, it is `System`. One `System` category can reference another `System` category. And we need readable output too.
```
Illegal reference to a 'Local' category 'Category' from Action.Category,Illegal reference to a 'Local' category 'Category' from Category.Master,Illegal reference to a 'Local' category 'Catalog' from Catalog.Parent,Illegal reference to a 'Local' category 'Category' from Identifier.Category,Illegal reference to a 'Local' category 'Language' from Localization.Language,Illegal reference to a 'Local' category 'Session' from Locking.Session,Illegal reference to a 'Local' category 'SystemUser' from Locking.SystemUser,Illegal reference to a 'Local' category 'SystemUser' from Log.SystemUser,Illegal reference to a 'Local' category 'Identifier' from Log.Identifier,Illegal reference to a 'Local' category 'Role' from Permission.Role,Illegal reference to a 'Local' category 'Category' from Permission.Category,Illegal reference to a 'Local' category 'Identifier' from Permission.Bound,Illegal reference to a 'Local' category 'Action' from Permission.Actions,Illegal reference to a 'Local' category 'System' from Replication.Source,Illegal reference to a 'Local' category 'System' from Replication.Destination,Illegal reference to a 'Local' category 'Category' from Replication.Category,Illegal reference to a 'Local' category 'Category' from Schema.Category,Illegal reference to a 'Local' category 'Subdivision' from Subdivision.Parent,Illegal reference to a 'Local' category 'Role' from SystemUser.Roles
```

PR #131, opened 2018-10-21
Fix invalid arguments on link verification

nechaido:
Fixes: https://github.com/metarhia/metaschema/issues/130

### DIFF
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -166,11 +166,12 @@ const linkSchemas = schemas => {
       } else if (field.category) {
         const category = categories.get(field.category);
         if (category) {
+          const categoryDefinition = category.definition;
           const error = verifyLink(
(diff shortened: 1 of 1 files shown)


lib/schema.js
```diff
@@ -164,7 +164,7 @@ const linkSchemas = schemas => {
           ));
         }
       } else if (field.category) {
-        const category = categories.get(field.category);
+        const category = categories.get(field.category).definition;
```
o-rumiantsev: If there will be no such category in `categories` this will throw an error. Maybe it is better to do like this
```javascript
if (category) {
  const { definition } = category;
  ...
}
```

nechaido:
Landed in:  1287b4cd9b79954aa452d6a99bcd416dfd260974.

Issue #132, opened 2018-10-22
Write tests for schema validation


nechaido:
We need tests that will check that validation passes for valid schemas and generates valid errors otherwise.

PR #134, opened 2018-10-22
Add tests for schema validation

o-rumiantsev:
Closes: https://github.com/metarhia/metaschema/issues/132

### DIFF
--- a/test/validateSchema.js
+++ b/test/validateSchema.js
@@ -1,0 +1,69 @@
+'use strict';
+
+const path = require('path');
+const metatests = require('metatests');
+const metaschema = require('..');
+
+const schemaDir = name => path.join(__dirname, 'schemas', name);
+
+const test = config => metatests.test(config.name, test => {
+  const dir = schemaDir(config.schemaName);
+  metaschema.fs.loadAndCreate(dir, null, error => {
+    if (!config.error) {
+      test.error(error);
+    } else {
+      test.isError(error);
+      test.equal(error.errors.length, 1, 'must only return one error');
+      const [actual] = error.errors;
+      test.strictSame(actual.type, config.error.type);
+      test.strictSame(actual.source, config.error.source);
+      test.strictSame(actual.property, config.error.property);
+    }
+    test.end();
+  });
+});
+
+const testConfigs = [
+  {
+    name: 'Validate unresolved domain',
+    schemaName: 'unresolvedDomainValidation',
+    error: {
+      type: 'unresolvedDomain',
+      source: 'UnresolvedDomainEntity',
+      property: 'UnresolvedDomain',
+    },
+  },
+  {
+    name: 'Validate unresolved category',
+    schemaName: 'unresolvedCategoryValidation',
+    error: {
+      type: 'unresolvedCategory',
+      source: 'UnresolvedCategoryEntity',
+      property: 'UnresolvedCategory',
+    },
+  },
+  {
+    name: 'Validate link to \'Log\' category',
+    schemaName: 'linkToLogValidation',
+    error: {
+      type: 'linkToLog',
+      source: 'LinkToLogEntity',
+      property: 'LinkToLog',
+    },
+  },
+  {
+    name: 'Validate illegal link to \'Local\'',
+    schemaName: 'illegalLinkToLocalValidation',
+    error: {
+      type: 'illegalLinkToLocal',
+      source: 'IllegalLinkToLocalEntity',
+      property: 'IllegalLinkToLocal',
+    },
+  },
+  {
+    name: 'Validate valid schemas',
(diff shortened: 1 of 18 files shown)


test/schemas/illegalLinkToLocalValidation/Local.schema
```diff
@@ -0,0 +1,5 @@
+{
```
tshemsedinov: Better naming: `LocalEntity.schema`

o-rumiantsev:
@tshemsedinov can a category be referenced by itself? Because now it will throw no error if we do like this:
```javascript
// CategoryName.schema
{
  FieldName: { category: 'CategoryName' }
}
```

belochub:
@o-rumiantsev, yes category may reference itself, this can be used to store graph-like (or tree-like as a special case) data structures. Take, for example, [Catalog](https://github.com/metarhia/globalstorage/blob/master/schemas/system/Catalog.schema) category from `globalstorage`.

test/validateSchema.js
```diff
@@ -0,0 +1,68 @@
...
+const schemaDir = name => path.join(__dirname, 'schemas', name);
+
+const test = config => metatests.test(config.name, test => {
+  const dir = schemaDir(config.schemaName);
+  metaschema.load(dir, errors => {
+    if (!config.error) {
+      test.error(errors[0], 'must not return errors');
+    } else {
+      test.assert(errors, 'must return errors');
+      test.equal(errors.length, 1, 'must return 1 error');
```
nechaido: `must only return one error` would sound better.

test/validateSchema.js
```diff
@@ -0,0 +1,72 @@
...
+      test.error(errors, 'must not return errors');
+    } else {
+      test.assert(errors, 'must return errors');
+      test.equal(errors.length, 1, 'must only return one error');
+      const type = config.error.type;
+      const source = config.error.source;
+      const property = config.error.property;
+      const [error] = errors;
+      const valid = error.type === type &&
+        error.source === source && error.property === property;
```
tshemsedinov: I'd prefer following formatting:
```js
const valid = (
  error.type === type &&
  error.source === source &&
  error.property === property
);
```

nechaido:
Blocked until https://github.com/metarhia/metaschema/pull/141.

test/structure.js
```diff
@@ -1,61 +1,66 @@
...
-      { x: 3, y: 5 },
+const geometryTest = metatests.test('geometry');
+const geometryPath = path.join(__dirname, '..', 'schemas', 'geometry');
+metaschema.fs.load(geometryPath, null, (err, ms) => {
+  geometryTest.error(err);
+
+  metatests.case('Metaschema / structure', {
+    createInstance: ms.createInstance.bind(ms),
+    buildCategory: ms.buildCategory.bind(ms),
+  }, {
```
tshemsedinov: Here we test API with separate functions, so do we need `bind`? I think this can work without `bind`.

test/structure.js
```diff
@@ -1,61 +1,66 @@
...
+    'createInstance': [
+      ['Point', { x: 3, y: 5 }, { x: 3, y: 5 }],
+      ['Point', { x: 3, y: 5 },        { x: 3, y: 5 }],
+      ['Point', [3, 5],                { x: 3, y: 5 }],
```
tshemsedinov: Strange code formatting/spacing

lundibundi:
@o-rumiantsev https://github.com/metarhia/metaschema/pull/141 has landed, you can now drop the commits from my branch and rebase on master.

test/validateSchema.js
```diff
@@ -0,0 +1,75 @@
...
+      const type = config.error.type;
+      const source = config.error.source;
+      const property = config.error.property;
+      const [actual] = error.errors;
+      const valid = (
+        actual.type === type &&
+        actual.source === source &&
+        actual.property === property
+      );
+      test.assert(valid, 'must return appropriate error');
```
lundibundi: Why not have 3 asserts instead of manually creating `valid`?
i.e.
```js
test.strictSame(actual.type, config.error.type);
test.strictSame(actual.source, config.error.source);
test.strictSame(actual.property, config.error.property);
```

lundibundi:
Also, have you considered just having every test case as a schema object in a file (instead of creating a separate file for each of them)? (e.g. https://github.com/metarhia/metaschema/blob/master/test/schema.js#L17)

o-rumiantsev:
@lundibundi I think it is easier to understand the logic of the test when schemas are in separate files.

belochub:
Landed in https://github.com/metarhia/metaschema/commit/c43f24c2b58a06dcaa83223e258a705bce7d0952.

PR #135, opened 2018-10-24
Add Action decorator

tshemsedinov:


lib/decorators.js
```diff
@@ -57,9 +57,10 @@ class Index {
 
 class Unique extends Index {}
 
-// Validation function decorator
+// Function decorator
```
belochub: Here and below:
```suggestion
// Function decorators
```

tshemsedinov:
Landed in 862e6de

PR #141, opened 2018-10-29
Refactor schema.js

lundibundi:
* remove global state usage -> Metaschema class with state
* split the implementation into separate parts (fs, schema, errors)

Most of the tests are still missing as there are PRs https://github.com/metarhia/metaschema/pull/134 and https://github.com/metarhia/metaschema/pull/111 so I'm not sure if I should write those myself or we will just rebase those PR against this branch.


lib/decorators.js
```diff
@@ -21,6 +21,13 @@ class Flags extends ValuesDecorator {
 const createFlags = (...values) => new Flags(values);
 createFlags.of = enumDomain => new Flags(null, enumDomain);
 
+class List extends ValuesDecorator {
```
nechaido: I'd prefer to make it a standalone class or create a new base class `CategoryDecorator` because it is a bit counterintuitive to store `category` in `this.values`.

test/structure.js
```diff
@@ -1,61 +1,66 @@
...
+          points: [
+            { x: 1, y: 3 },
+            { x: 2, y: 2 },
+            { x: 3, y: 4 },
+            { x: 4, y: 5 },
+            { x: 3, y: 5 },
+          ],
+        },
+      ],
+      */
```
lundibundi: This is to be uncommented when @tshemsedinov fixes `List` decorator. (in a follow up PR)

belochub: @lundibundi can you remove the following line from PR's description then:
>proper List decorator implementation

lib/schema-fs-loader.js
```diff
@@ -0,0 +1,98 @@
...
+    filename,
+    timeout: SCRIPT_TIMEOUT,
+    lineOffset: 1,
+  };
+  return vm.runInNewContext(code, sandbox, options);
+};
+
+// Load schema file from the filesystem
+//   filepath - <string> full path to the schema file
+//   callback - <Function> to report the result
```
belochub: `api` argument's description missing.

lib/schema.js
```diff
@@ -1,122 +1,27 @@
...
-//   destination - object
-//   destinationName - string
-//   propertyName - string
-// Returns: object, information about error or null if link is valid
+//   source - <Object>
+//   sourceName - <string>
+//   destination - <Object>
+//   destinationName - <string>
+//   propertyName - <string>
+// Returns: <Object> | <null> information about error or null if link is valid
```
```suggestion
// Returns: <ValidationError> | <null> information about error or null if link is valid
```

lib/schema-fs-loader.js
```diff
@@ -0,0 +1,112 @@
...
+    lineOffset: 1,
+  };
+  return vm.runInNewContext(code, sandbox, options);
+};
+
+// Load schema file from the filesystem
+//   filepath - <string> full path to the schema file
+//   api - <Object> object to be passed as 'api' during loading
+//   callback - <Function> to report the result
+//     error - <Error> error or null
```
SemenchenkoVitaliy:
```suggestion
//     error - <Error> | <null>
```

lib/schema-fs-loader.js
```diff
@@ -0,0 +1,112 @@
...
+//     name - <string> schema name
```
SemenchenkoVitaliy: To be correctly parsed, there should be a comma after types.
```suggestion
//     name - <string>, schema name
```

lib/schema-fs-loader.js
```diff
@@ -0,0 +1,112 @@
...
+      callback(e);
+      return;
+    }
+    callback(null, name, exports);
+  });
+};
+
+const loadFiles = (files, api, callback) => {
+  const schemas = [];
+  const domainsPath = path.join(schemasDir, 'metaschema', DOMAINS_SCHEMA);
```
aqrln: Should `'metaschema'` be a named constant too?

lib/schema.js
```diff
@@ -186,230 +94,194 @@ const linkSchemas = schemas => {
...
     }
-    if (domain && type !== domain.type) {
-      result.errors.push(`Field ${key}:${val} expected to be ${domain.type}`);
+    return errors.length === 0 ? null : new MetaschemaError(errors);
+  }
+
+  // Validate fields of an instance against a category
+  //   category - <string> category name
+  //   instance - <any> instance to validate
+  // Returns: <Object> result
```
```suggestion
  // Returns: <MetaschemaError> | <null>
```

test/structure.js
```diff
@@ -1,61 +1,66 @@
...
-      { x: 3, y: 5 },
+const geometryTest = metatests.test('geometry');
+const geometryPath = path.join(__dirname, '..', 'schemas', 'geometry');
+metaschema.fs.loadAndCreate(geometryPath, null, (err, ms) => {
+  geometryTest.error(err);
+
+  metatests.case('Metaschema / structure', {
+    createInstance: ms.createInstance.bind(ms),
+    buildCategory: ms.buildCategory.bind(ms),
+  }, {
```
tshemsedinov: Here we test API with separate functions, so do we need `bind`? I think this can work without `bind`.

test/structure.js
```diff
@@ -1,61 +1,66 @@
...
+    'createInstance': [
+      ['Point', { x: 3, y: 5 }, { x: 3, y: 5 }],
+      ['Point', { x: 3, y: 5 },        { x: 3, y: 5 }],
+      ['Point', [3, 5],                { x: 3, y: 5 }],
```
tshemsedinov: Strange formatting/spacing

test/structure.js
lundibundi: I'm not sure I understand, createInstance and buildCategory use 'this' inside, how will this work without bind?

lib/schema-fs-loader.js
lundibundi: As I asked in a comment to the refactoring PR, could we avoid this (https://github.com/metarhia/common/pull/194#issuecomment-433872681)?

lib/schema-fs-loader.js
lundibundi: I extracted whole domainsPath and inlined the variables, wdyt?

test/structure.js
tshemsedinov: Understood, you use `this` context to hold schema. Maybe will be better to hold it in closure, but now this is a test of instance, so we have syntax for that.

PR #147, opened 2018-11-03
Prepare for browser (expose decorators)

lundibundi:


lib/schema-converters.js
```diff
@@ -0,0 +1,32 @@
+'use strict';
+
+const { iter } = require('metarhia-common');
+const { all: decorators } = require('./decorators');
+
+const defaultOptions = {
+  useEsm: false,
+};
```
tshemsedinov: Maybe to a single line

lib/schema-converters.js
```diff
@@ -0,0 +1,32 @@
...
+
+const createImport = name => `import { ${name} } from 'metaschema';`;
+const createRequire = name => `const { ${name} } = require('metaschema');`;
+
+const generateImports = (usedDecorators, useEsm) =>
+  // eslint-disable-next-line implicit-arrow-linebreak
+  `${(useEsm ? createImport : createRequire)('decorators')}\n` +
+  `const { ${usedDecorators.join(', ')} } = decorators;`;
```
tshemsedinov: Maybe use round brackets with multi-line expression
```js
const generateImports = (usedDecorators, useEsm) => (
  `${(useEsm ? createImport : createRequire)('decorators')}\n` +
  `const { ${usedDecorators.join(', ')} } = decorators;`
);
```

lib/schema-converters.js
```diff
@@ -0,0 +1,32 @@
...
+    .toArray();
+  return (usedDecorators.length > 0 ?
+    generateImports(usedDecorators, useEsm) + '\n\n' : '') +
+    `${useEsm ? 'export default' : 'module.exports ='} ${source};`;
+};
+
+
+module.exports = {
+  toJs,
+};
```
tshemsedinov: Maybe to a single line

test/converters.js
```diff
@@ -0,0 +1,66 @@
...
+    const expected = `export default ${simpleSchema};`;
+    t.strictSame(toJs(simpleSchema, { useEsm: true }), expected);
+  });
+
+  test.testSync('must convert decorated schema', t => {
+    const expected =
+    `const { decorators } = require('metaschema');
+const { Many, Include } = decorators;
+
+module.exports = ${decoratedSchema};`;
```
tshemsedinov: I do not like multi-line templates

lib/schema-converters.js
```diff
@@ -0,0 +1,32 @@
...
+const generateImports = (usedDecorators, useEsm) =>
+  // eslint-disable-next-line implicit-arrow-linebreak
+  `${(useEsm ? createImport : createRequire)('decorators')}\n` +
+  `const { ${usedDecorators.join(', ')} } = decorators;`;
+
+const toJs = (source, options = {}) => {
+  const { useEsm } = Object.assign({}, defaultOptions, options);
+  const usedDecorators = iter(Object.keys(decorators))
+    // eslint-disable-next-line no-useless-escape
+    .filter(name => source.match(`${name}\(.*\)`))
```
aqrln: The ESLint warning is legit and I don't quite understand why you disabled it since it can't give any false-positives. These should be `\\(` and `\\)`.

lib/schema-converters.js
lundibundi: This is the style we have for all other such options/defaultOptions and don't want to have it differently here.

lib/schema-converters.js
```diff
@@ -0,0 +1,31 @@
...
+
+const generateImports = (usedDecorators, useEsm) =>
+  // eslint-disable-next-line implicit-arrow-linebreak
+  `${(useEsm ? createImport : createRequire)('decorators')}\n` +
+  `const { ${usedDecorators.join(', ')} } = decorators;`;
+
+const toJs = (source, options = {}) => {
+  const { useEsm } = Object.assign({}, defaultOptions, options);
+  const usedDecorators = iter(Object.keys(decorators))
+    .filter(name => source.match(`${name}\\([\\S\\s]*\\)`))
```
aqrln: This doesn't account for possible whitespace between the identifier and the left paren, and for possible identifier part to the left of `${name}`. I'd just match for `\\b${name}\\b`, but if you think that checking for parens is absolutely necessary, it should be something like `\\b${name}\\s*(`.

lib/schema-converters.js
tshemsedinov: Ok

test/converters.js
lundibundi: IMO it makes it easier to understand/edit this as it is intended to be converted/saved in a file.

tshemsedinov: Ok

belochub:
Landed in https://github.com/metarhia/metaschema/commit/6323925b0a54ad31f833aa06679c086b9514734d.

PR #150, opened 2018-11-06
Build the package for usage with frontend module bundlers

aqrln:


package.json
```diff
@@ -21,11 +21,17 @@
   "bin": {
     "metaschema": "./bin/cli.js"
   },
+  "browser": {
+    "./metaschema.js": "./dist/metaschema.js",
+    "./dist/lib/schema-fs-loader.js": false
```
belochub: I think It will also make sense to set `./dist/lib/generator.js` and `./dist/lib/introspection.js` to false as well; I am not sure how they can be used in the browser.

PR #151, opened 2018-11-07
Allow to pass multiple directories to loadAndCreate

lundibundi:


belochub:
Landed in https://github.com/metarhia/metaschema/commit/d8ab530fb9ac09529437bfa2755db739a52948b4.

PR #155, opened 2018-11-08
Fix invalid comments

belochub:
We've missed this during the review of https://github.com/metarhia/metaschema/pull/141.


belochub:
Landed in https://github.com/metarhia/metaschema/commit/07c747eab473f654a9ba6dbb470af6752e6062fc.

PR #156, opened 2018-11-12
Return and store sources of loaded schemas

lundibundi:
* return source as 3rd element of an array from `load`
* store schema sources as `Metaschema.sources`
* accept 'source' as 3rd element of elements in the argument to `create`


lib/schema-fs-loader.js
```diff
@@ -33,6 +33,7 @@ const processSchemaFile = (filename, decorators, source, api = {}) => {
 //     error - <Error> | <null>
 //     name - <string> schema name
 //     schema - <Object> parsed schema
+//     source - <String> schema source
```
belochub:
```suggestion
//     source - <string> schema source
```

PR #157, opened 2018-11-13
Store Actions in Metaschema 

nechaido:


lib/schema-utils.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+const extractDecorator = schema => {
+  const className = schema.constructor.name;
+  if (className) {
+    return className;
+  }
+
+  return Object.getPrototypeOf(schema).name;
+};
```
tshemsedinov:
```js
const extractDecorator = schema =>
  schema.constructor.name || Object.getPrototypeOf(schema).name;
```

tshemsedinov: And I thin we can put this function to `lib/schema.js`

lib/schema.js
```diff
@@ -269,18 +274,45 @@ class Metaschema {
...
     ms[addSchema](name, schema, source);
   }
-  return [linkSchemas(schemas, ms.categories, ms.domains), ms];
+
+  const error = linkSchemas(schemas, ms.categories, ms.domains);
+  if (error) {
+    return [error, ms];
+  }
+
+  ms[initActions]();
```
tshemsedinov: What it initActions?

nechaido: `initActions` is a Symbol, so that is is impossible to call this method outside of the module.

metaschema.js
```diff
@@ -16,5 +16,6 @@ module.exports = Object.assign(
     fs: schemaFs,
     create,
     decorators: decorators.all,
+    extractDecorator,
```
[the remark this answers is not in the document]
nechaido: So that it can be used in other modules that rely on metaschema (i.e. globalstorage).

lib/schema.js
```diff
@@ -269,21 +283,49 @@ class Metaschema {
...
+      );
+    }
+  }
 }
 
 // Creates Metaschema instance
 //   schemas - <Iterable> schemas in form [name, schema, source]
 //             (the 'source' is optional)
-// Returns: <Metaschema>
+// Returns: [<MetaschemaError>, <Metaschema>]
```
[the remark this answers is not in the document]
nechaido: It is used in `loadAndCreate` to return an error in a callback, moreover, it is an established behavior, I am merely fixing a comment.

lib/schema.js
```diff
@@ -1,17 +1,33 @@
...
+//   schema - <Object>
+// Returns: <string>
+const extractDecorator = schema => {
+  const className = schema.constructor.name;
+  if (className !== 'Function') {
+    return className;
+  }
+
+  const functionDecorator = Object.getPrototypeOf(schema).name;
+  return functionDecorator ? functionDecorator : 'Function';
```
```suggestion
  return functionDecorator || 'Function';
```

lib/schema.js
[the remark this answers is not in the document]
belochub: Okay then, I was just confused with the changes to this function below. :smile:

lib/schema.js
```diff
@@ -269,21 +286,49 @@ class Metaschema {
...
     ms[addSchema](name, schema, source);
   }
-  return [linkSchemas(schemas, ms.categories, ms.domains), ms];
+
+  const error = linkSchemas(schemas, ms.categories, ms.domains);
+  if (error) {
+    return [error, ms];
+  }
+
+  ms[initActions]();
```
belochub: Can't you call this method before the call to `linkSchemas()`?
It will make this function much simpler:
```js
  const ms = new Metaschema();
  for (const [name, schema, source] of schemas) {
    ms[addSchema](name, schema, source);
  }
  ms[initActions]();
  return [linkSchemas(schemas, ms.categories, ms.domains), ms];
```

lib/schema.js
```diff
@@ -269,21 +285,43 @@ class Metaschema {
   buildCategory(category, ...args) {
     return this.categories.get(category).factory(...args);
   }
+
+  [initActions]() {
+    for (const [categoryName, { definition: category }] of this.categories) {
+      this.actions.set(
+        categoryName,
+        iter(Object.keys(category))
```
lundibundi: Nit: I'd prefer to have this chain outside of `.set` call and have assigned a separate variable.

lib/schema.js
```diff
@@ -269,21 +285,43 @@ class Metaschema {
...
+      );
+    }
+  }
 }
 
 // Creates Metaschema instance
 //   schemas - <Iterable> schemas in form [name, schema, source]
 //             (the 'source' is optional)
-// Returns: <Metaschema>
+// Returns: [<MetaschemaError>, <Metaschema>]
```
lundibundi: I don't think we support such notation though =).

nechaido: Well, we probably should.

lib/schema.js
nechaido: Resolved.

lib/schema.js
nechaido: IIUC we will as soon as https://github.com/metarhia/metaschema/pull/107 lands.

belochub: @SemenchenkoVitaliy can you confirm that such notation will be supported?
Also, what about:
```js
// Returns: <Array>
//   <MetaschemaError>
//   <Metaschema>
```
Will it be possible to write it that way?

SemenchenkoVitaliy: ### Creates Metaschema instance
`create(schemas)`
  - `schemas` - `<Iterable>` schemas in form [name, schema, source]
        (the 'source' is optional)

Returns: `[ <MetaschemaError>, <Metaschema> ]`

This is the part of generated md file with corrected comments.

> Also, what about:
> 
> ```js
> // Returns: <Array>
> //   <MetaschemaError>
> //   <Metaschema>
> ```
> 
> Will it be possible to write it that way?

Yes, this is totally possible and correct, though we should choose one so we will have same comments style across all files. @tshemsedinov @belochub @nechaido @lundibundi WDYT?

lib/schema.js
```diff
@@ -1,17 +1,32 @@
 'use strict';
 
+const { iter } = require('metarhia-common');
+
 const { ValidationError, MetaschemaError } = require('./schema-errors');
 
 const DOMAINS_NAME = 'domains';
 const addSchema = Symbol('addSchema');
 const factorify = Symbol('factorify');
+const initActions = Symbol('initAction');
```
```suggestion
const initActions = Symbol('initActions');
```

lundibundi:
Landed in edaa12ef97a22b8fd614f1afd5ccd58fb360a7a0 and be9a02dbd9b5c4d07659345a2d0b7b08f25fb27a.

Issue #98, opened 2018-09-03
Schema support for non floating numbers


nechaido:
Currently, we don't provide a way to define whether a number is an integer or not, also I would be also nice to define desired precision for the number.

I propose to add 2 Domain attributes: `precision` and `floating`.

tshemsedinov:
One attribute is enough, for example: `DomainName: { type: 'number', precision: 'real/double' }` Using pgslq `NUMERIC(precision, scale)` isn't good practice for numbers except `Money` type.

lundibundi:
I'd suggest at least having some enum of types:
`Enum('i32', 'i64', 'u32', 'u64', 'f32', 'f64')` or something,

tshemsedinov:
Why `i8` and `i16` are not included?

lundibundi:
I thought that they are not really useful in a general use case (and I'm not sure if it's possible to properly map them in DB schema generation), but we can as well have them.

tshemsedinov:
Pgsql have `smallint` (i16), `integer` (i32) and `bigint` (i64). It have no `tinyint` type (i8) that may be implemented by `char(1)` or `create domain tinyint as smallint constraint tiny check (value >= 0 and value <= 255);`

nechaido:
@nechaido we should include arbitrary precision numbers to store numbers without precision loss, also having `precision` and `scale` rules for any number is useful for validating user input (values that should have less than a specific amount of fractional digits).

PR #158, opened 2018-11-16
Updates to domain and field schemas

tshemsedinov:


lundibundi:
Though I'd prefer to have more granular control over floating values than just int/double (as noted by @nechaido in https://github.com/metarhia/metaschema/issues/98#issuecomment-439069862).

schemas/metaschema/Domain.schema
```diff
@@ -13,7 +12,9 @@
...
   min: { type: 'number' },
   max: { type: 'number' },
-  signed: { type: 'boolean' },
   bytes: { type: 'number' },
-  comment: { type: 'string' }
+  comment: { type: 'string' },
+  parse: { type: 'function' },
+  serialize: { type: 'function' },
+  normalize: { type: 'function' },
 }
```
nechaido: Add validation function as well?

tshemsedinov: We have `validate` field in `DatabaseField` but domains may not be linked to record, so do not support calculation and validation because we have no place in code to dynamically execute this schema. I'll move `normalize` to `DatabaseField` too @nechaido

nechaido: @tshemsedinov it can be used when we validate an object against a category, form or anything else, see https://github.com/metarhia/metaschema/pull/154/files#diff-f412f4722a52cf22ed3d0a9a694b837dR187.

schemas/metaschema/DatabaseField.schema
```diff
@@ -1,17 +1,18 @@
 {
   domain: { domain: 'Nomen' },
   category: { domain: 'Nomen' },
-  required: { domain: 'Logical' },
+  required: { domain: 'Logical' }, // empty string is not considered as correct
```
nechaido: It may be better to add a `truthy` property to a record or treat `min` for `string` as a minimum length.

tshemsedinov: Main purpose for schemas is "to be subject domain oriented", so `truthy: true` is clear for developer, but `required: true` is better for any analyst.

schemas/metaschema/Domain.schema
```diff
@@ -16,4 +16,6 @@
   comment: { type: 'string' },
   parse: { type: 'function' },
   serialize: { type: 'function' },
+  check: { type: 'function' }, // (value) => (Boolean)
+  normalize: { type: 'function' }, // (value, record, category) => (value)
```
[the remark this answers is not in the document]
tshemsedinov: Agree, fixed

schemas/metaschema/DatabaseField.schema
nechaido: What about using `min` for string, because an empty string is still a string? I don't like the idea of `require` working differently throughout types: the initial value for numbers (`0`) is valid, yet the initial value for strings (`''`) is not.

tshemsedinov: Required will deny only: `undefined`, `null`, `not set`, `NaN`

schemas/metaschema/Domain.schema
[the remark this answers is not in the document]
belochub: @tshemsedinov and what about record value? You haven't removed it.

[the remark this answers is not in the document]
tshemsedinov: `record` is an `object` containing this field, if applicable. We may swap arguments: `(value, domain, record) => (value)`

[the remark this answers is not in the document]
belochub: +1 to swapping the arguments.

tshemsedinov:
Landed in 714f695

PR #170, opened 2018-12-04
Save references to a category

nechaido:


lib/schema.js
```diff
@@ -93,24 +93,36 @@ const linkSchemas = (schemas, categories, domains) => {
...
+        if (type === 'Include') {
+          references = category.includedBy;
+        } else if (type === 'Many') {
+          references = category.manyReferences;
+        } else if (type === 'Master') {
+          references = category.masterOf;
+        } else {
+          references = category.referencedBy;
         }
+        references.push({ category: schemaName, property: fieldName });
```
lundibundi: I think it'll be easier to have references in a separate object of references separated by decorator name. i.e.
```js
const referenceTypes = new Set(['Include', 'Many', 'Master']);
...
let type = extractDecorator(field);
if (!referenceTypes.has(type)) type = 'other';
category.references[type].push({ category: schemaName, property: fieldName }))
```
(Also, we can put `'other'` or any other name in the referenceTypes to simplify the creation of the initial object `referenceTypes.forEach(t => { category.references[t] = []; })`)

IMO this will simplify their management, make them easier to find (you don't have to remember 4 new names for the fields) and group them better in general.

lib/schema.js
```diff
@@ -93,24 +93,36 @@ const linkSchemas = (schemas, categories, domains) => {
...
+        const error = verifyLink(
+          schema, schemaName, categoryDefinition, field.category, fieldName
+        );
+        if (error) {
+          errors.push(error);
+          continue;
+        }
+        field.definition = categoryDefinition;
+        let references;
+        const type =  extractDecorator(field);
```
```suggestion
        const type = extractDecorator(field);
```

PR #172, opened 2018-12-05
Allow properties not defined in schema

nechaido:


tshemsedinov:
We need special syntax in schema to allow such fields. For example:
```js
{
  FieldName: { domain: 'DomainName' },
  Symbol('other'): true
}
```

nechaido:
@tshemsedinov int that case we would need to manually add `Id` into every category definition.

PR #173, opened 2018-12-06
Extract View and Projection definition

nechaido:


lib/schema.js
```diff
@@ -129,6 +129,48 @@ const linkSchemas = (schemas, categories, domains) => {
...
+    for (const key in props) {
+      const prop = props[key];
+      if (typeof prop !== 'string') {
+        view.definition[key] = prop;
+        continue;
+      }
+      let schema = ms.categories.get(view.Category);
+      if (!schema) {
+        schema = ms.views.get(view.Category);
+      }
```
lundibundi: Nit: 
```js
const schema =
 ms.categories.get(view.Category) || ms.views.get(view.Category);
```

test/schemas/view/FullName.schema
```diff
@@ -0,0 +1,4 @@
+{
+    FirstName: { domain: 'Nomen' },
+    LastName: { domain: 'Nomen' },
+}
```
lundibundi: I think the indentation in schemas should be 2, isn't it?

lib/schema.js
```diff
@@ -129,6 +129,48 @@ const linkSchemas = (schemas, categories, domains) => {
...
+};
+
+const processView = (view, ms) => {
+  if (view.definition) {
+    return;
+  }
+  view.definition = {};
+  for (const field of view.Fields) {
+    const props = typeof field === 'string' ? { [field]: field } : field;
+    for (const key in props) {
```
lundibundi: Nit: perhaps `Object.keys(props)[0]` to make sure we only support one key there or signify that there shouldn't be more than one field.

nechaido: This is intentionally implemented this way to support:
```js
View({
  Category: 'Name',
  Fields: [{
    A: 'B',
    B: 'A',
  }],
})
```

test/schemas/view/FullName.schema
nechaido: AFAIK it's not specified anywhere, we have no linting rules for schemas,  and indentation has no effect on it's parsing.

belochub:
Landed in https://github.com/metarhia/metaschema/commit/21a4e0a1ae50267958777dc7a7d23203d3af9db1.

PR #174, opened 2018-12-07
Add DataStructure.schema, DatabaseCategory.schema

tshemsedinov:
Refs: https://github.com/metarhia/metaschema/pull/172

### DIFF
--- a/schemas/metaschema/DataStructure.schema
+++ b/schemas/metaschema/DataStructure.schema
@@ -1,0 +1,3 @@
+{
+  Symbol('allowNotDefinedFields'): false
+}
--- a/schemas/metaschema/DataStructure.schema
+++ b/schemas/metaschema/DataStructure.schema
@@ -1,3 +1,3 @@
 {
-  Symbol('allowNotDefinedFields'): false
+  Symbol.for('AllowNotDefinedFields'): false
 }
--- a/schemas/metaschema/DatabaseCategory.schema
+++ b/schemas/metaschema/DatabaseCategory.schema
@@ -1,4 +1,4 @@
 {
   Id: { domain: 'Id', required: true },
-  Symbol('allowNotDefinedFields'): false
+  Symbol.for('AllowNotDefinedFields'): false
 }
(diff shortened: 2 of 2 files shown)


schemas/metaschema/DatabaseCategory.schema
```diff
@@ -0,0 +1,4 @@
+{
+  Id: { domain: 'Id', required: true },
+  Symbol('allowNotDefinedFields'): false
```
lundibundi: Perhaps this was meant to be `Symbol.for`? As `Symbol` will make it impossible to actually get this field IIUC.
Also, I'd suggest to 'namespace' it `@metarhia/allowNotDefinedFields`.

schemas/metaschema/DataStructure.schema
```diff
@@ -0,0 +1,3 @@
+{
+  Symbol.for('AllowNotDefinedFields'): false
```
belochub: That's an invalid syntax, you have to enclose the computed property name into the brackets:
```suggestion
  [Symbol.for('AllowNotDefinedFields')]: false
```

schemas/metaschema/DatabaseCategory.schema
```diff
@@ -0,0 +1,4 @@
+{
+  Id: { domain: 'Id', required: true },
+  Symbol.for('AllowNotDefinedFields'): false
```
belochub: Ditto.
```suggestion
  [Symbol.for('AllowNotDefinedFields')]: false
```

belochub:
It seems that it was correctly failing on Node.js 6 due to the invalid syntax in the schema definitions, but I am not sure why it hasn't failed on other Node.js versions. This may be caused by a bug in either this package or the `metatests` package.

nechaido:
> @nechaido this field specifies whether not defined fields are allowed or no

@tshemsedinov, I see the purpose of the symbol, yet I can't get how the schemas themselves should be used.

nechaido:
@tshemsedinov ping, see: https://github.com/metarhia/metaschema/pull/174#issuecomment-447256269.

nechaido:
Closed in favor of #221.

PR #175, opened 2018-12-11
Refactor Action implementation

lundibundi:
Add collection domain

* make Action an object with properties
* add Execute decorator to denote Action reference
* add type, transactionMode, processMode, refreshMode props
* allow to specify custom Form name via Form field

### DIFF
--- a/lib/decorators.js
+++ b/lib/decorators.js
@@ -96,7 +96,6 @@ class Action {
     } else {
       Object.assign(this, def);
     }
-    if (!this.Type) this.Type = 'Custom';
     if (!this.TransactionMode) this.TransactionMode = 'All';
     if (!this.ProcessMode) this.ProcessMode = 'Any';
     if (!this.RefreshMode) this.RefreshMode = 'Current';
--- a/lib/decorators.js
+++ b/lib/decorators.js
@@ -96,10 +96,10 @@ class Action {
     } else {
       Object.assign(this, def);
     }
-    if (!this.type) this.type = 'Custom';
-    if (!this.transactionMode) this.transactionMode = 'All';
-    if (!this.processMode) this.processMode = 'Any';
-    if (!this.refreshMode) this.refreshMode = 'None';
+    if (!this.Type) this.Type = 'Custom';
+    if (!this.TransactionMode) this.TransactionMode = 'All';
+    if (!this.ProcessMode) this.ProcessMode = 'Any';
+    if (!this.RefreshMode) this.RefreshMode = 'None';
   }
 }
 
--- a/lib/decorators.js
+++ b/lib/decorators.js
@@ -99,7 +99,7 @@ class Action {
     if (!this.Type) this.Type = 'Custom';
     if (!this.TransactionMode) this.TransactionMode = 'All';
     if (!this.ProcessMode) this.ProcessMode = 'Any';
-    if (!this.RefreshMode) this.RefreshMode = 'None';
+    if (!this.RefreshMode) this.RefreshMode = 'Current';
   }
 }
 
--- a/lib/decorators.js
+++ b/lib/decorators.js
@@ -78,7 +78,30 @@ class Unique extends Index {}
 // Function decorators
 
 class Validate {}
-class Action {}
+
+class Execute {
+  constructor(def) {
+    if (typeof def === 'string') {
+      this.Action = def;
+    } else {
+      Object.assign(this, def);
+    }
+  }
+}
+
+class Action {
+  constructor(def) {
+    if (typeof def === 'function') {
+      this.Execute = def;
+    } else {
+      Object.assign(this, def);
+    }
+    if (!this.type) this.type = 'Custom';
+    if (!this.transactionMode) this.transactionMode = 'All';
+    if (!this.processMode) this.processMode = 'Any';
+    if (!this.refreshMode) this.refreshMode = 'None';
+  }
+}
 
 // Entity decorators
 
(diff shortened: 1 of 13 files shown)


tshemsedinov:
FYI: @aqrln we need to know context specification (schema) to understand how to implement this PR better. Example (from your paper):
```js
context.schema: {
  Id: { domain: 'Id', required: false, comment: 'current record id' },
  SelectedIds: { domain: 'List', required: false, comment: 'records selected in ui' },
  Parent: { domain: 'Id', required: false, comment: 'direct parent id' },
  Master: { domain: 'Id', required: false, comment: 'high-level id' },
  Category: { domain: 'Nomen', required: true, comment: 'category name' },
  ParentCategory: { domain: 'Nomen', required: false },
  MasterCategory: { domain: 'Nomen', required: false },
  ProcessId: { required: false, comment: 'business-process id' },
  DisplayMode: { domain: 'Nomen', required: false, comment: 'old name show method' },
  DisplayModeParameters: { domain: 'Collection', required: false },
}
```

lib/decorators.js
```diff
@@ -78,7 +78,30 @@ class Unique extends Index {}
...
+}
+
+class Action {
+  constructor(def) {
+    if (typeof def === 'function') {
+      this.Execute = def;
+    } else {
+      Object.assign(this, def);
+    }
+    if (!this.type) this.type = 'Custom';
```
tshemsedinov:
```suggestion
    if (!this.Type) this.Type = 'Custom';
```

lib/decorators.js
```diff
@@ -78,7 +78,30 @@ class Unique extends Index {}
...
+    if (!this.transactionMode) this.transactionMode = 'All';
```
tshemsedinov:
```suggestion
    if (!this.TransactionMode) this.TransactionMode = 'All';
```

lib/decorators.js
```diff
@@ -78,7 +78,30 @@ class Unique extends Index {}
...
+    if (!this.processMode) this.processMode = 'Any';
```
tshemsedinov:
```suggestion
    if (!this.ProcessMode) this.ProcessMode = 'Any';
```

lib/decorators.js
```diff
@@ -78,7 +78,30 @@ class Unique extends Index {}
...
+    if (!this.refreshMode) this.refreshMode = 'None';
```
tshemsedinov:
```suggestion
    if (!this.RefreshMode) this.RefreshMode = 'None';
```

schemas/metaschema/domains.schema
```diff
@@ -66,4 +67,19 @@
   Volume: { type: 'number', unit: 'Litre', symbol: 'l' },
   Speed: { type: 'number', unit: 'Metres per second', symbol: 'm/s' },
   Ratio: { type: 'number' },
+  ActionType: Enum(
+    'Create', 'Update', 'Delete', 'ShowDetail',
+    'Digicert', 'Steps', 'Custom'
```
belochub: What's the `Steps` action type, how will it behave?

lundibundi: As of now it's not really specified how will `steps` work.
It describes an Action that may happen via multiple interactions with the user (and corresponding executions of different Actions) while keeping some internal state.

There is no use for it now so I can remove it, though we will probably need to add it later anyway.

belochub: Doesn't this PR add an `Execute` decorator, that is supposed to determine the next step in the interaction with the user?
Or did I misunderstand the purpose of it and there is a significant difference?

lundibundi: Yes, this is a basic idea - to return the next step via Execute, but that may not be enough when we actually get to implementing such an interaction.

tshemsedinov: `'Create', 'Update', 'Delete'` is ok, but we don't need `'ShowDetail', 'Digicert', 'Steps', 'Custom'` @lundibundi

schemas/metaschema/domains.schema
```diff
@@ -66,4 +67,19 @@
...
+  ),
+  // determines how Action with multiple items will be contained
+  // Each - commit after each item, All - commit after all
+  TransactionMode: Enum('Each', 'All'),
```
tshemsedinov:
```suggestion
  TransactionMode: Enum('Single', 'All'),
```

tshemsedinov: `All` by default. By the way if `Action` is `Sequential` or `Parallel` we will disallow to return `Execute` for multiple entities.

schemas/metaschema/domains.schema
```diff
@@ -66,4 +67,19 @@
...
+  // determines how Action uses selected items
+  // Any - may use any elements, Current - uses only current element
+  // Selected - uses all user selected items
+  ProcessMode: Enum('Any', 'Current', 'Selected'),
+  // determines what rows to update after the Action
+  // None - don't update, Current - update only affected items
+  // All - update whole dataset
+  RefreshMode: Enum('None', 'Current', 'All'),
```
tshemsedinov:
```suggestion
  RefreshMode: Enum('Current', 'All', 'None'), // Current by defeult
```

lundibundi:
Landed in https://github.com/metarhia/metaschema/commit/b446ecc939523cf86c6fd0ab7ad9969edacb563f and https://github.com/metarhia/metaschema/commit/ed310873b3c78464405634720c52280ee8d288e3.

PR #177, opened 2018-12-12
Use `Execute('Action')` instead of multiple methods

tshemsedinov:


lundibundi:
Merged in https://github.com/metarhia/metaschema/pull/175.

Issue #171, opened 2018-12-04
ApplicationMenu description


nechaido:
We need a way to define a Menu in an application schema. The proposed structure is:
```js
MenuItem = {
  type: Enum('caterory', 'projection', 'group', 'separator'),
  name: { domain: 'Nomen' },
  icon: { domain: 'Nomen', comment: 'resource key for menu entry icon' },
  items: Array('MenuItem'),
};

ApplicationMenu = Array('MenuItem');
```

lundibundi:
I'd propose:
```js
MenuItem = {
  type: Enum('category', 'projection', 'group', 'separator'),
  name: { domain: 'text' },
  icon: { domain: 'text', comment: 'resource key for menu entry icon' },
  items: List('MenuItem'),
};

ApplicationMenu = List('MenuItem');
```

Fixed a few typos.
I think 'Nomen' is way too limited, so switched to text.
I don't think we have 'Array' decorator, 'List'?

igorgo:
The `name`  prop is a name of category or projection, isn`t? Аnd where is the most important thing. Where is the **internationalized** text to display?

nechaido:
@igorgo 
> The name prop is a name of category or projection, isn`t? 

Yes, it is.

As per internationalized string, I believe that resources should be stored somewhere else, not in the schema.

PR #178, opened 2018-12-12
Add forms processing

nechaido:


test/schemas/form/Person.ChangeDOB.schema
```diff
@@ -0,0 +1,27 @@
+Form({
+  Args: {
+    OldDOB: { field: 'DOB' },
+    NewDOB: { domain: 'DateTime' },
+    LastName: { field: 'FullName.LastName' },
+  },
+  Markup: {
```
tshemsedinov:
```suggestion
  Layout: {
```

test/schemas/form/Person.ChangeDOB.schema
```diff
@@ -0,0 +1,27 @@
...
+    Tabs: {
+      control: 'tabs',
+      children: {
+        tab: {
+          control: 'tab',
+          children: {
+            group: {
+              control: 'group',
```
tshemsedinov: We need real control names from UI group @alinkedd

test/schemas/form/Person.schema
```diff
@@ -0,0 +1,5 @@
+{
+  DOB: { domain: 'DateTime' },
+  FullName: { category: 'FullName' },
+  Mother: { category: 'Person' },
```
tshemsedinov: Where is a `Father`? :smile:

test/schemas/form/Person.ChangeDOB.schema
```diff
@@ -0,0 +1,27 @@
...
+              children: {
+                OldDOB: { control: 'date' },
+                NewDOB: { control: 'date' },
```
tshemsedinov: Why do we have OldDOB? If this control will be just for info, we need a way to make it disabled.

test/schemas/validate/Schema1.Form.schema
```diff
@@ -0,0 +1,5 @@
+Form({
+  Args: {
```
tshemsedinov: What is `Args` for forms?

test/schemas/form/Person.schema
igorgo: :trollface: `Mother` and `Father` is discrimination. According to the last directions, person may have two mothers :two_women_holding_hands: or tho fathers :two_men_holding_hands:

test/schemas/form/Person.ChangeDOB.schema
igorgo: It seems to me that this already goes far beyond the `metaschema`. Forms are a part of the `MIA framework`

igorgo: The same can be said about #171

nechaido: IIUC it is not, metaschema only provides a way to load schemas, processing of a schema may or may not be proprietary, but if we intend to load something by metaschema, we need to add a way to load it and test that it is loaded correctly.

nechaido: @tshemsedinov 
> We need real control names from UI group

Form Layout is not a subject of this PR, any data contained here is just an example of a structure we agreed upon so that `Layout` or `Markup` props are not an empty object.

test/schemas/form/Person.ChangeDOB.schema
nechaido: Same as https://github.com/metarhia/metaschema/pull/178/files#r241332710.

test/form.js
```diff
@@ -0,0 +1,25 @@
+'use strict';
+
+const metatests = require('metatests');
+
+const metaschema = require('..');
+const { getSchemaDir } = require('./utils');
+
+metatests.test('must load Views', test => {
```
belochub: You don't seem to load any `View`s in this test despite the name.

test/schemas/form/Person.ChangeDOB.schema
tshemsedinov: Ok, it's just a stub

test/schemas/validate/Schema1.Form.schema
tshemsedinov: I think all fields should be passed to an action. Why need no `Fields` and `Args`

nechaido: There is no way to define arguments on an action, it was decided previously that arguments to an `Action`s should be defined on a corresponding `Form`.

aqrln:
Instead of relying on the order of keys of objects, it would be both more correct and more convenient to parse it if we moved control names to other properties and made `children` arrays. Syntactically, those arrays might be just arguments of control decorators:

```js
Layout: [
  Group({ name: 'tabs', control: 'Tabs' },
    Group({ name: 'tab', control: 'Tab' },
      Group({ name: 'group' },
        Input({ name: 'OldDOB' }),
        Input({ name: 'NewDOB' }),
        Input({ name: 'LastName' }),
      )
    )
  )
]
```

We only need a minimal set of decorators that specify families of controls (like groups and inputs), and refine the control type further given the domain. However, the type may be specified explicitly to override the defaults.

To make the syntax even shorter, @igorgo suggested to pass control name as the first argument of the corresponding decorator:

```js
Layout: [
  Group('tabs', { control: 'Tabs' },
    Group('tab', { control: 'Tab' },
      Group('group',
        Input('OldDOB'),
        Input('NewDOB'),
        Input('LastName'),
      )
    )
  )
]
```

igorgo:
Since there can be separate labels I suggest to add the `Label` decorator instead of using the `Input` one with `{control: 'Label'}`

lib/schema.js
```diff
@@ -129,21 +129,95 @@ const linkSchemas = (schemas, categories, domains) => {
...
+};
+
+const extractByPath = (definition, path, ms, source, property) => {
+  const isOuterCategory = path.startsWith('::');
+  if (!isOuterCategory) {
+    return extractDefinition(definition, path, ms, '', source, property);
+  }
+
+  const separatorIndex = path.indexOf('.');
+  if (separatorIndex === -1) {
```
lundibundi: I think Ditto.

lib/schema.js
```diff
@@ -157,18 +231,82 @@ const processView = (view, ms) => {
...
+        errors.push(new SchemaValidationError(
+          'unresolvedDomain',
+          source,
+          `${prop}.${key}`,
+          { domain: field.domain }
+        ));
+      } else {
+        Object.assign(field, { domain: field.domain, definition: def });
+      }
+      continue;
```
lundibundi: Nit: I think `else` would look better here.

Edit: Perhaps the second argument to the `Object.assign` can be extracted too and the call to it will be moved to the end of the function.

lib/schema.js
```diff
@@ -616,9 +785,20 @@ const create = schemas => {
...
     ms[addSchema](name, schema, source);
   }
-  ms[initActions]();
-  processViews(ms);
-  return [linkSchemas(schemas, ms.categories, ms.domains), ms];
+  const error = linkSchemas(schemas, ms.categories, ms.domains);
+  if (error) {
+    return [error, ms];
+  }
+  const actionIntiError = ms[initActions]();
```
```suggestion
  const actionInitError = ms[initActions]();
```

test/paramsValidation.js
```diff
@@ -0,0 +1,118 @@
...
+    test.isError(error, expected);
+
+    const errors = error.errors.map(error => {
+      delete error.stack;
+      return error;
+    });
+
+    const expectedErrors = expected.errors.map(error => {
+      delete error.stack;
+      return error;
```
lundibundi: Nit: I think this will look better as just `forEach`.

lib/schema.js
```diff
@@ -538,18 +676,31 @@ class Metaschema {
   }
 
   // Validate instance against a category
-  //   formName - <string>, form name
+  //   category - <string>
+  //   action - <string>
   //   args - <Object>, form arguments
```
```suggestion
   //   args - <Object>, action arguments
```

lib/schema.js
```diff
@@ -538,18 +676,31 @@ class Metaschema {
   }
 
   // Validate instance against a category
```
```suggestion
   // Validate action arguments
```

lib/schema.js
```diff
@@ -157,18 +231,82 @@ const processView = (view, ms) => {
...
+    }
+  }
+
+  return errors;
+};
+
+const processForms = ms => {
+  const errors = [];
+  for (const [name, form] of ms.forms) {
+    const dot = name.indexOf('.');
```
belochub: Should this maybe also return an error when there is no `.` symbol in `Form`'s name?

lib/schema.js
```diff
@@ -129,21 +129,75 @@ const linkSchemas = (schemas, categories, domains) => {
...
+  if (!isOuterCategory) {
+    return extractDefinition(definition, path, ms, '', info);
+  }
+
+  const separatorIndex = path.indexOf('.');
+  const categoryName =
+    separatorIndex === -1 ? path.slice(2) : path.slice(2, separatorIndex);
+  const category = ms.categories.get(categoryName);
+
+  const endOfCategory = separatorIndex === -1 ? path.length : separatorIndex;
```
lundibundi: Why not
```js
  const separatorIndex = path.indexOf('.');
  const endOfCategory = separatorIndex === -1 ? path.length : separatorIndex;
  const categoryName = path.slice(2, endOfCategory);
  const category = ms.categories.get(categoryName);
```
?

lib/schema.js
```diff
@@ -129,21 +129,75 @@ const linkSchemas = (schemas, categories, domains) => {
...
+    throw new SchemaValidationError(
+      'unresolvedCategory',
+      info.source,
+      info.property,
+      { category: path.slice(0, endOfCategory) }
+    );
+  }
+
+  return separatorIndex === -1 ?
+    { category: categoryName, definition: category.definition } :
```
belochub: Can't you just return `category` here?

nechaido: Because it doesn't have `category` property.

lib/schema.js
```diff
@@ -537,19 +668,32 @@ class Metaschema {
     return errors.length === 0 ? null : new MetaschemaError(errors);
   }
 
-  // Validate instance against a category
-  //   formName - <string>, form name
-  //   args - <Object>, form arguments
+  // Validate arguments on an action
```
```suggestion
  // Validate arguments of an action
```

nechaido:
Nit: use object spread instead of `Object.assign`.

Issue #184, opened 2019-01-09
Use object spread instead of `Object.assign`


nechaido:
Refs: https://github.com/metarhia/metaschema/pull/178#issuecomment-452753248

PR #187, opened 2019-01-10
Use object spread instead of Object.assign

nechaido:


lib/decorators.js
```diff
@@ -126,7 +126,7 @@ class Arguments extends Decorator {}
 
 class LayoutDecorator {
   constructor(name, config = {}) {
-    Object.assign(this, config, { name });
+    Object.assign(this, { ...config, name });
```
tshemsedinov: I'd prefer previous solution

belochub:
Fixes: https://github.com/metarhia/metaschema/issues/184

PR #188, opened 2019-01-10
Use async functions for Actions

nechaido:
Closes: https://github.com/metarhia/metaschema/issues/183

### DIFF
--- a/test/schema-actions.js
+++ b/test/schema-actions.js
@@ -9,9 +9,11 @@ const { getSchemaDir } = require('./utils');
 metatests.test('must load Actions', test => {
 
   const testMethod = (test, method, ctx, expected) => {
-    method({}, ctx, {}, (err, actual) => {
-      test.error(err);
+    method({}, ctx, {}).then(actual => {
       test.strictSame(actual, expected);
+      test.end();
+    }, err => {
+      test.error(err);
       test.end();
     });
   };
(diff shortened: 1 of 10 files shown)


test/schemas/actionParamsValidation/Person.schema
```diff
@@ -3,7 +3,7 @@
   FullName: { category: 'FullName' },
   Mother: { category: 'Person' },
   ChangeDOB: Action({
-    Execute: (session, ctx, args, callback) => callback(null, 'ok'),
+    Execute: async (session, ctx, args, callback) => 'ok',
```
belochub: You should remove the `callback` argument here. This also applies to other `Action`s.
```suggestion
    Execute: async (session, ctx, args) => 'ok',
```

test/schema-actions.js
```diff
@@ -9,7 +9,7 @@ const { getSchemaDir } = require('./utils');
 metatests.test('must load Actions', test => {
 
   const testMethod = (test, method, ctx, expected) => {
-    method({}, ctx, {}, (err, actual) => {
+    method({}, ctx, {}).then((actual, err) => {
```
belochub: `Promise.prototype.then()` passes only one argument to the `onFulfilled` function (first argument to `then()`). You should add a second argument (`onRejected` function) to `then()` here, to handle error.

test/schema-actions.js
```diff
@@ -91,21 +93,25 @@ metatests.test('must load Actions', test => {
...
       const t1 = test.test();
-      execute({}, { Id: 42 }, {}, (err, act) => {
-        t1.error(err);
+      execute({}, { Id: 42 }, {}).then(act => {
         t1.type(act, 'Execute');
         t1.strictSame(act.Action, 'M1');
         t1.strictSame(act.Form, 'CustomForm');
         t1.end();
+      }, err => {
+        test.error(err);
```
belochub: Why do you use `test` and not `t1` that was used before here?

PR #189, opened 2019-01-10
Expose processSchema to allow loading of 'string' schemas

lundibundi:


lib/schema-fs-loader.js
```diff
@@ -43,7 +28,7 @@ const loadSchema = (filepath, context, callback) => {
     const name = path.basename(filepath, '.schema');
     let exports;
     try {
-      exports = processSchemaFile(filepath, decorators, source, context);
+      exports = processSchema(name, source, context);
```
belochub: Should we maybe keep the `decorators` argument in case anyone wants to specify custom decorators? It can also be optional.

lib/schema-loader.js
```diff
@@ -0,0 +1,29 @@
...
+
+const vm = require('vm');
+const { all: decorators } = require('./decorators');
+
+const SCRIPT_TIMEOUT = 2000;
+
+// Process schema source
+//   name - <string> schema name
+//   source - <string> schema source
+//   context - <Object> object to be assigned to global during loading
```
belochub: Missing description for `evaluator` argument.

lib/schema-loader.js
```diff
@@ -0,0 +1,29 @@
...
+// Returns: <Object> processed schema
+const processSchema = (
+  name, source, context = {}, evaluator = vm.runInNewContext
```
belochub: Is there a need to make `evaluator` configurable?

lundibundi: In order to allow to use this without polyfill in browser.
/cc @aqrln

lib/schema-fs-loader.js
lundibundi: At this point I don't see any reason to have it (or need to specify custom decorators), I think we can easily add it back if we ever need it.

lib/schema-loader.js
belochub: Oh, OK then, I just thought we were going to use `webpack` to polyfill `vm` for browsers.

lib/schema-fs-loader.js
belochub: It will enable the usage of the decorators declared outside of this library by the users of this package.
This can also help with the frequent release of new versions of this package whenever we add or change decorators here.

lib/schema-loader.js
```diff
@@ -0,0 +1,37 @@
...
+//   context - <Object> object to be assigned to global during loading
+//   evaluator - <Function> function to be used to evaluate schema
+//   decorators - <Object> decorators available during schema processing
+// Returns: <Object> processed schema
+const processSchema = (
+  name, source, context = {}, evaluator = null, decorators = allDecorators
+) => {
+  if (!evaluator) {
+    // In order to natively support browser don't require 'vm' until requested
+    if (!vm) vm = require('vm');
```
nechaido: I don't like this statement.

PR #192, opened 2019-01-14
Use prettier for code formatting

JuliaGerasymenko:


package.json
```diff
@@ -33,7 +33,8 @@
   ],
   "scripts": {
     "test": "npm run lint && metatests test/",
-    "lint": "eslint --ignore-path .gitignore .",
+    "lint": "eslint . --ignore-path .gitignore",
```
nechaido: Is this needed?

JuliaGerasymenko: I found that in most of repositories the dot goes after eslint not at the end of a command. Alright, I'll change it back.

README.md
```diff
@@ -17,32 +18,28 @@ $ npm install metaschema
 
 #### create(schemas)
 
-  - `schemas: `[`<Iterable>`] schemas in form [name, schema, source] (the
-        'source' is optional)
-
-*Returns:* `[ `[`<MetaschemaError>`]`, `[`<Metaschema>`]` ]`
+- `schemas:`[`<Iterable>`] schemas in form [name, schema, source] (the
+  'source' is optional)
```
SemenchenkoVitaliy: I would prefer that this line had this line had the same indentation from parameter comment beginning.

nechaido:
@JuliaGerasymenko please add `README.md` to `.prettierignore`, don't forget to reset changes i `README.md`.

JuliaGerasymenko:
@nechaido Should I do the same in another repositories?

nechaido:
@JuliaGerasymenko this should be done for any autogenerated file.

.prettierignore
```diff
@@ -0,0 +1,5 @@
+node_modules
```
belochub: There is no need to add `node_modules` here since it is ignored by default.

.prettierignore
```diff
@@ -0,0 +1,5 @@
...
+*.log
```
belochub: These files are not `.js`, `.md` or `.json`, and you run `prettier` only on files with extensions mentioned above:
https://github.com/metarhia/metaschema/blob/2256c955ad58d93374f15ffd0b104b7d5e650e0d/package.json#L36-L37
So there is no point in adding this line.
```suggestion
```

.prettierignore
```diff
@@ -0,0 +1,5 @@
...
+.DS_Store
```
belochub: Same here.
```suggestion
```

package.json
```diff
@@ -33,7 +33,8 @@
   ],
   "scripts": {
     "test": "npm run lint && metatests test/",
-    "lint": "eslint --ignore-path .gitignore .",
+    "lint": "eslint --ignore-path .gitignore . && prettier --debug-check \"**/*.js\" \"**/*.json\" \"**/*.md\" .eslintrc .prettierrc  --ignore-path .prettierignore",
```
nechaido: `.prettierignore` is used by default
```suggestion
    "lint": "eslint --ignore-path .gitignore . && prettier --debug-check \"**/*.js\" \"**/*.json\" \"**/*.md\" .eslintrc .prettierrc",
```

package.json
```diff
@@ -33,7 +33,8 @@
...
+    "fmt": "prettier --write \"**/*.js\" \"**/*.json\" \"**/*.md\" .eslintrc .prettierrc --ignore-path .prettierignore",
```
nechaido:
```suggestion
    "fmt": "prettier --write \"**/*.js\" \"**/*.json\" \"**/*.md\" .eslintrc .prettierrc",
```

PR #197, opened 2019-01-23
Make metaschema loadable in browser

nechaido:


lib/schema.js
```diff
@@ -500,33 +500,118 @@ const addCategoryData = (ms, schemas) => {
...
+          definition: value,
+        });
+      }
+    });
+  }
+};
+
+const typeToProps = {
+  category: 'categories',
+  domains: 'domains',
```
aqrln: Is it intentional that the name on the left hand side is in plural too?

nechaido: Yes, because there may be many domains in every object.

PR #201, opened 2019-01-24
Treat datasets in DisplayModes as optional

nechaido:


test/schemas/displayModes/Category.Mode.display
```diff
@@ -0,0 +1,4 @@
+DisplayMode({
```
tshemsedinov: `Category.display` not `Category.Mode.display`

nechaido: `Category.sidplay` is not valid, because it is located in root directory of schemas, thus it should have the form: `categoryName.displayModeName.display`.

lib/schema-processors.js
```diff
@@ -271,6 +271,9 @@ const processActions = ms => {
 
 const processDisplayMode = (display, category, ms) => {
   const errors = [];
+  if (!display.definition.datasets) {
```
belochub: I think it'd be maybe better to move this check to `processDisplayModes()` function to avoid unnecessary calling this function and creating and pushing an empty array.

lundibundi: I'm -1 on this because this check only makes sense in something that deals with the DisplayMode abstraction itself but `processDisplayModes` is only supposed to know about the collection/separation of `DisplayMode`s and not their internals. The fact that we have nothing to do (yet) if the `datasets` is empty may easily change in the future (and probably will) therefore requiring us to move this check back again to this function.

test/schemas/displayModes/Category.Mode.display
lundibundi: I'd prefer a more explicit schema name though. Perhaps `Category.EmptyMode.display` or `Category.NoDatasetsMode.display`.

PR #206, opened 2019-01-30
Decouple metaschema from predefined schemas

nechaido:
@belochub @lundibundi @tshemsedinov PTAL at proposed function signatures: `create` and `load`.

The current proposal is that the processing of schemas is done by 4 functions:
* `preprocess` - change\add properties to a schema, that was omitted to reduce boilerplate, e.g. `category` in `Hierarchical`;
* `validate` - validate schema;
* `add` - save schema into a specific field in a `Metaschema` object;
* `postprocess` - cross-link and cross-validate schemas.

Existing processors would be exported from the package as default proposed configuration.


lib/fs-loader.js
```diff
@@ -0,0 +1,132 @@
...
+
+const readdir = util.promisify(fs.readdir);
+const readFile = util.promisify(fs.readFile);
+const stat = util.promisify(fs.stat);
+
+const flatten = arr =>
+  arr.reduce((acumulator, element) => {
+    acumulator.push(...element);
+    return acumulator;
+  }, []);
```
lundibundi: `[].concat(...arr)`?

lib/config.js
```diff
@@ -0,0 +1,136 @@
...
+        postprocess: [postprocessDomains],
+        validator: validate.domains,
+      },
+      category: {
+        add: [addCategory],
+        postprocess: [postprocessCategory],
+        validator: validate.category,
+      },
+    },
+    order,
```
lundibundi: I'd suggest renaming this to `processOrder` and the other `order` to `loadOrder` to make the difference between them clear IIUC.

lib/fs-loader.js
```diff
@@ -0,0 +1,132 @@
...
+//           Returns: <Error[]>
+//         create <Function>, optional
+//           ms <Metaschema>
+//           schema <Schema>
+//           args <any>
+//           options <Object>
+//           Returns:
+//             <Error[]>
+//             <any>
+//     order: <Function>, function to pass to Array.prototype.sort,
```
lundibundi: Could we perhaps allow this to be an object of sort `{ 'type1': 0, 'type2': 1 }` etc and transform it into a sorting function. As this will be the main use case IMO.

lib/errors.js
```diff
@@ -1,49 +1,42 @@
...
-  constructor(type, source, property, info) {
+  constructor(type, source, info) {
     super();
     this.type = type;
     this.source = source;
+    this.info = info;
+  }
+
+  toString() {
+    return SchemaValidationError.Serializers[this.type](this);
```
belochub:
```suggestion
    return SchemaValidationError.serializers[this.type](this);
```

lib/errors.js
```diff
@@ -1,49 +1,42 @@
...
+  }
+}
+
+SchemaValidationError.serizlizers = {
```
belochub:
```suggestion
SchemaValidationError.serializers = {
```

lib/errors.js
```diff
@@ -1,49 +1,42 @@
...
-    return schemaValidationErrorSerializers[this.type](this);
+    return ValidationError.serializers[this.type](this);
   }
 }
 
 const wrap = value =>
   Array.isArray(value) ? value.map(wrap).join(', ') : `'${value}'`;
 
-const validationErrorSerializers = {
+ValidationError.serizlizers = {
```
belochub:
```suggestion
ValidationError.serializers = {
```

lib/config.js
```diff
@@ -0,0 +1,135 @@
...
+
+const getLoadOrder = name => {
+  if (!name.includes('.')) {
+    return 2;
+  } else if (name.endsWith('.category')) {
+    return 1;
+  } else {
+    return 0;
+  }
+};
```
tshemsedinov:
```js
const getLoadOrder = name => {
  if (!name.includes('.')) return 2;
  if (name.endsWith('.category')) return 1;
  return 0;
};
```

PR #208, opened 2019-02-11
Add schemas preprocessor

nechaido:


tshemsedinov:
What is a purpose of schemas preprocessor?

nechaido:
@tshemsedinov in case we need to change the array of schemas. For example, we need to extract all `actions` from a category.

Issue #212, opened 2019-02-15
Increase test coverage


nechaido:
We need to find parts of code not covered by tests and add a corresponding test for those cases.

PR #213, opened 2019-02-19
Increase test coverage

o-rumiantsev:
Closes: https://github.com/metarhia/metaschema/issues/212

### DIFF
--- a/test/default.unit.js
+++ b/test/default.unit.js
@@ -1,0 +1,72 @@
+'use strict';
+
+const metatests = require('metatests');
+
+const {
+  Metaschema,
+  default: { decorators, options, config },
+} = require('..');
+
+const { SchemaValidationError } = require('../lib/errors');
+
+metatests.testSync('default config adders', test => {
+  const ms = new Metaschema(config);
+
+  config.prepare(ms);
+
+  const [addDomains] = config.processors.domains.add;
+  addDomains({ definition: { Nomen: { type: 'string' } } }, ms);
+
+  test.strictSame(ms.domains.get('Nomen'), { type: 'string' });
+
+  test.strictSame(
+    addDomains({ definition: { Nomen: { type: 'string' } } }, ms),
+    [
+      new SchemaValidationError('duplicate', 'Nomen', null, {
+        entity: 'domain',
+      }),
+    ]
+  );
+
+  const [addCategory] = config.processors.category.add;
+  addCategory({ name: 'Person' }, ms);
+
+  test.strictSame(ms.categories.get('Person'), { name: 'Person' });
+
+  test.strictSame(addCategory({ name: 'Person' }, ms), [
+    new SchemaValidationError('duplicate', 'Person', {
+      type: 'category',
+      value: 'Person',
+    }),
+  ]);
+
+  test.strictSame(options.pathToType('name.domains'), 'domains');
+  test.strictSame(options.pathToType('name.category'), 'category');
+
+  test.strictSame(config.processOrder, {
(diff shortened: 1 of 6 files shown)


test/default.js
```diff
@@ -70,5 +83,54 @@ metatests.test('must properly process schemas', async test => {
     test.strictSame(category, categories[name]);
   }
 
+  const [addDomains] = config.processors.domains.add;
```
nechaido: This looks like a unit test, you don't need to load schemas from filesystem to check that `add` works the way it is supposed to work.

test/default.js
```diff
@@ -70,5 +83,54 @@ metatests.test('must properly process schemas', async test => {
...
+  test.strictSame(
+    addDomains({ definition: { Nomen: { type: 'string' } } }, ms),
```
nechaido: Ditto.

test/default.js
```diff
@@ -70,5 +83,54 @@ metatests.test('must properly process schemas', async test => {
...
+    [
+      new SchemaValidationError('duplicate', 'Nomen', null, {
+        entity: 'domain',
+      }),
+    ]
+  );
+
+  const [addCategory] = config.processors.category.add;
```
nechaido: Ditto.

test/default.unit.js
```diff
@@ -0,0 +1,68 @@
...
+
+const { SchemaValidationError } = require('../lib/errors');
+
+metatests.testSync('default config adders', test => {
+  const ms = new Metaschema(config);
+
+  config.prepare(ms);
+
+  const [addDomains] = config.processors.domains.add;
+  addDomains({ definition: { Nomen: { type: 'string' } } }, ms);
```
nechaido: You should check that it is actually added, same for a category.

test/errors.js
```diff
@@ -0,0 +1,99 @@
+'use strict';
+
+const metatests = require('metatests');
+const { SchemaValidationError, ValidationError } = require('../lib/errors.js');
+
+const errorsTest = metatests.test('errors test');
+
+errorsTest.testSync('SchemaValidationError', test => {
+  const unresolved = new SchemaValidationError('unresolved', 'source', {
```
nechaido:
```js
[
  {
    args: ['unresolved', 'source', { type: 'type', value: 'value' }],
    expected: `Reference to unresolved type 'value' from 'source'`,
  },
  ...
].forEach(({args, expected) => {
  const error = new SchemaValidationError(...args);  
  test.strictSame(error.toString, expected);
})
```

Same for `ValidationError`.

test/schema-processor.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+const vm = require('vm');
+const metatests = require('metatests');
+const { processSchema } = require('../lib/schema-processor');
+
+const processorTest = metatests.test('schema processor test');
+processorTest.plan(1);
```
nechaido: This looks like a synchronous test, use 'testSync'.

test/schema.js
```diff
@@ -0,0 +1,95 @@
...
+      type: 'type',
+      name: 'name',
+      module: 'module',
+      source: 'source',
+    },
+  ];
+
+  ms.schemas.push(...sources);
+
+  test.strictSame(ms.sources, sources);
```
nechaido: This test looks like it a test that `Array.prototype.push` works.

nechaido:
Please rebase this onto master.

Note: we have changed the way we handle errors, they are now thrown as a single instance of `MetaschemaError` instead of returning as an array.

PR #215, opened 2019-02-21
Improve error handling

belochub:
Avoid returning the array of errors from the public methods and instead
return null or a single instance of Error that contains one or more
instances of Error in it. Also, reject a promise returned from
`metaschema.fs.load()` on errors during `Metaschema` object creation
instead of resolving it with an array of errors.

### DIFF
--- a/lib/fs-loader.js
+++ b/lib/fs-loader.js
@@ -127,7 +127,7 @@ const loadModule = async (dir, options) => {
 //         If <Object> is provided it would be used as
 //         map from schema type (<string>) to order value (<number>),
 //         types with lower values are processed earlier.
-// Returns: [<Error[]>, <Metaschema>]
+// Returns: <Metaschema>
 const load = async (dir, options, config) => {
   if (!Array.isArray(dir)) {
     dir = [dir];
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const { getProcessOrder } = require('./utils');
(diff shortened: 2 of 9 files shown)


test/unresolved.js
```diff
@@ -13,15 +13,12 @@ const { getSchemaDir } = require('./utils');
...
 
   try {
-    [errors, ms] = await load(path, options, config);
-  } catch (error) {
-    test.fail(error);
-    test.end();
-    return;
+    await load(path, options, config);
+  } catch (err) {
+    error = err;
```
nechaido: I'd check that `err instanceof MetaschemaError`.

lib/fs-loader.js
```diff
@@ -140,7 +140,9 @@ const load = async (dir, options, config) => {
   const schemas = options.preprocessor
     ? options.preprocessor(flatten(results))
     : flatten(results);
-  return Metaschema.create(schemas, config);
+  const [err, ms] = Metaschema.create(schemas, config);
```
nechaido: I'd prefer to throw an error from `Metaschema.create`.

lib/schema.js
```diff
@@ -38,7 +39,11 @@ class Metaschema {
     if (!validator) {
       throw new TypeError(`No validator defined for type: '${type}'`);
     }
-    return validator(this, schema, instance, options);
+    const errors = validator(this, schema, instance, options);
+    if (errors.length === 0) {
```
lundibundi: Why not use a wrapMetaschemaError here?

PR #223, opened 2019-03-11
Version 0.3.3

nechaido:


nechaido:
Landed in 285d86a.

tshemsedinov:
Why don't update lodash too?

belochub:
@tshemsedinov, this package doesn't directly depend on `lodash`, so there is no way to update it explicitly here.

PR #228, opened 2019-04-10
Implement list decorator

bugagashenkj:


lib/validate.js
```diff
@@ -120,6 +120,24 @@ const validateCategory = (
...
+    const opts = { ...options };
+
+    let err;
+    if (definition.domain) {
+      opts.path = `${path}${prop}`;
+      err = ms.validate('domains', definition.domain, value, opts);
+    } else if (definition.category) {
+      opts.path = `${path}${prop}.`;
+      err = ms.validate('category', definition.category, value, opts);
+    }
```
o-rumiantsev: `opts.path` is repeated in both cases

```javascript
const opts = { ...options };
opts.path = `${path}${prop}`;

let err;
if (definition.domain) {
  err = ms.validate('domains', definition.domain, value, opts);
} else if (definition.category) {
  err = ms.validate('category', definition.category, value, opts);
}
```

lib/validate.js
```diff
@@ -158,20 +176,16 @@ const validateCategory = (
       continue;
     }
 
-    const opts = { ...options };
+    if (extractDecorator(definition) === 'List') {
+      value.forEach((ins, index) => {
```
o-rumiantsev:
```suggestion
      value.forEach((listItem, index) => {
```

lib/validate.js
bugagashenkj: In the case of a domain, we have a dot at the end.

lib/validate.js
```diff
@@ -158,20 +176,16 @@ const validateCategory = (
...
-    if (definition.domain) {
-      opts.path = `${path}${prop}`;
-      err = ms.validate('domains', definition.domain, value, opts);
-    } else if (definition.category) {
-      opts.path = `${path}${prop}.`;
-      err = ms.validate('category', definition.category, value, opts);
-    }
-    if (err) {
-      errors.push(...err.errors);
+      continue;
```
nechaido: I'd prefer not to add `continue` here, so that `definition.validate` is checked.

lib/decorators.js
```diff
@@ -53,5 +62,6 @@ module.exports = {
     Flags: createFlags,
     // Function decorators
     Validate: fn => new Validate(fn),
+    List: createList,
```
nechaido: `List` should be category specific decorator.

PR #229, opened 2019-04-10
Fix fs.load modifing arguments

bugagashenkj:


lib/fs-loader.js
```diff
@@ -132,13 +132,20 @@ const load = async (dir, options, config) => {
   if (!Array.isArray(dir)) {
     dir = [dir];
   }
+  const newOptions = Object.assign({}, options);
```
o-rumiantsev:
```suggestion
  const newOptions = common.clone(options);
```

lundibundi:
```suggestion
  const options = { ...options };
```
?

Also, why is there a need to make copy?

test/default.unit.js
```diff
@@ -44,8 +44,10 @@ metatests.testSync('default config adders', test => {
     }),
   ]);
 
-  test.strictSame(options.pathToType('name.domains'), 'domains');
-  test.strictSame(options.pathToType('name.category'), 'category');
+  if (typeof options.pathToType === 'function') {
```
nechaido: AFAIU, options.pathToType is always an object.

PR #234, opened 2019-04-18
Cleanup documentation

o-rumiantsev:


lib/schema-processor.js
```diff
@@ -8,6 +8,8 @@ const SCRIPT_TIMEOUT = 2000;
 //   options <Object>
 //     context <Object> object to be assigned to global during loading
 //     decorators <Object> decorators available during schema processing
+//     localDecorators <Object> specific decorators for different
```
nechaido:
```suggestion
//     localDecorators <Object> type specific decorators
```

lib/fs-loader.js
```diff
@@ -70,60 +70,64 @@ const loadModule = async (dir, options) => {
...
+//     context <Object>
+//     decorators <Object>
 //     localDecorators <Object>, optional
 //       [type] <Object> decorators specific for this schema type
 //     pathToType <Object>
 //       [ext] <string>
 //     pathToType <Function>
 //       path <string>
 //     Returns: <string>
+//     loadOrder <Function>, optional
```
nechaido: There is no such argument.

lib/fs-loader.js
```diff
@@ -70,60 +70,64 @@ const loadModule = async (dir, options) => {
...
+//       filename1 <string>
+//       filename2 <string>
+//     Returns: <number>
 //     preprocessor: <Function>, optional
-//       schemas <Schema[]>
-//     Returns: <Schema[]>
+//       schemas <Object[]>
```
nechaido: `<Schema>` is intentional, there is no need to change it to `<Object>` anywhere.

lib/fs-loader.js
o-rumiantsev: https://github.com/metarhia/metaschema/blob/8162a20cb76fea0777926298c15667bedb4a047d/lib/config.js#L130

nechaido: Remove it, it's not used anywhere.

PR #235, opened 2019-04-23
Add Class field for Enum and Flags domains

bugagashenkj:


lib/config.js
```diff
@@ -11,6 +14,9 @@ const postprocessDomains = (domains, ms) => {
       const enumDomain = ms.domains.get(domain.enum);
       domain.values = enumDomain.values;
     }
+    if (decorator === 'Flags' || decorator === 'Enum') {
+      domain.Class = common[decorator].from(domain.values);
```
nechaido:
```suggestion
      domain.Class = common[decorator].from(...domain.values);
```

nechaido: You should also add a test, to ensure that classes are properly created.

nechaido:
You should also add a test, to ensure that classes are properly created.

PR #237, opened 2019-05-11
Version 0.3.5

nechaido:


nechaido:
This version is unpublished due to the fact that `Flags` decorator now throws if there are more than 64 values.

belochub:
@nechaido, it isn't unpublished: https://www.npmjs.com/package/metaschema/v/0.3.5.

PR #238, opened 2019-05-13
Handle Flags with more than 64 values

nechaido:


test/decorators.js
```diff
@@ -36,3 +36,17 @@ metatests.test('Decorators / Flags', async test => {
...
+    await load('test/schemas/decorators-invalid', options, config);
+    test.fail('must fail');
+  } catch (error) {
+    test.strictSame(error.errors.length, 1);
+    test.isError(
+      error.errors[0],
+      new TypeError('Flags does not support more than 64 values')
+    );
+  }
+  test.end();
```
SemenchenkoVitaliy: AFAIK, this is not necessary. /cc @lundibundi.
```suggestion
```

PR #379, opened 2022-02-11
Few fixes + validation function

lundibundi:
- [x] description of changes is added in CHANGELOG.md
- [x] update .d.ts typings


lib/schema.js
```diff
@@ -319,6 +326,16 @@ class Schema {
...
 
+function extractSchema(def) {
+  if (def instanceof Schema) {
+    return def;
+  }
+  if (def.schema instanceof Schema) {
+    return def.schema;
+  }
+  return null;
+}
```
tshemsedinov, line 337:
```suggestion
const extractSchema = (def) => {
  if (def instanceof Schema) return def;
  if (def.schema instanceof Schema) return def.schema;
  return null;
};
```

PR #391, opened 2022-03-31
WIP: Metaschema new features and refactoring

georgolden:
Done:
- [x] now support any nested type definitions with combination of reference types
- [x] now can be added any custom types
- [x] changed Model behavior - now types have different format
- [x] changed behavior Schema - now every reference represented by type `reference` to be able to validate and check data in runtime using only relational schemas
- [x] fixed issue with projection loading. Previously, if Projection was loaded before their parent - error occurred.
- [x] test `metasql` compatibility

To-do:
- [ ] add description to CHANGELOG.md
- [ ] move preprocess from Schema to a separate class Preprocessor
- [ ] update validate handling due to review
- [ ] update .d.ts typings
- [ ] prepare an update for `metasql` to support new `.types` interface
- [ ] create dictionary for converting types from `metachema` format to `typescript`, now only references added
- [ ] replace Kind with object to have potentially custom Kinds support and remove hardcoded projection


lib/model.js
```diff
@@ -3,8 +3,9 @@
 const fsp = require('fs').promises;
 const { Schema } = require('./schema.js');
 const { readDirectory } = require('./loader.js');
-const { TYPES } = require('./types.js');
+const { TYPES, prepareTypes } = require('./types.js');
 
+const last = new Set();
```
tshemsedinov, line 8:
```suggestion
const last = new Set();

```

lib/model.js
```diff
@@ -3,8 +3,9 @@
 const fsp = require('fs').promises;
 const { Schema } = require('./schema.js');
 const { readDirectory } = require('./loader.js');
-const { TYPES } = require('./types.js');
+const { TYPES, prepareTypes } = require('./types.js');
```
tshemsedinov, line 6: `prepareTypes` is not clear to me by its name what it may contain

lib/schema.js
```diff
@@ -13,119 +19,33 @@ const SCOPE = ['application', 'global', 'local'];
...
-    return Object.assign(def, { type, value });
-  } else if (type === 'map' || type === 'object') {
-    const col = def[type];
-    Reflect.deleteProperty(def, type);
-    const key = Object.keys(col)[0];
-    const value = col[key];
-    return Object.assign(def, { type, key, value });
-  } else {
-    return { type };
+const setValidate = (validate) => {
```
tshemsedinov, line 26: `setValidate` is also not clear naming

lib/schema.js
```diff
@@ -135,97 +55,117 @@ class Schema {
...
     this.relations = new Set();
-    this.validate = defs.validate || null;
-    this.format = defs.format || null;
-    this.parse = defs.parse || null;
-    this.serialize = defs.serialize || null;
-    this.preprocess(defs);
+    this.validate = setValidate(raw.validate);
+    this.format = raw.format || (() => {});
+    this.parse = raw.parse || (() => {});
+    this.serialize = raw.serialize || (() => {});
```
tshemsedinov, line 61: We don't need additional round parentheses around functions here

lib/schema.js
```diff
@@ -135,97 +55,117 @@ class Schema {
...
-    for (const ref of references) {
-      if (isFirstUpper(ref)) {
-        if (!this.findReference(ref)) {
-          warn.push(`Warning: "${ref}" referenced by "${name}" is not found`);
-        }
-      } else if (!this.findType(ref)) {
-        warn.push(`Warning: type "${ref}" is not found in "${name}"`);
-      }
+  findKind(name) {
+    const kind = KIND.find((el) => el === name);
```
tshemsedinov, line 162:
```suggestion
    const kind = KIND.includes(name) ? name : undefined;
```

lib/schema.js
```diff
@@ -239,84 +179,84 @@ class Schema {
...
-        else if (!isType) type = 'string';
+        let fieldName = name + (isEntity ? 'Id' : '');
+        if (!isFirstLetter(fieldName)) fieldName = `"${fieldName}"`;
+        if (isEntity) {
+          type = 'number';
+        } else if (!isType) {
+          const { array, reference, relation } = defs;
+          if (array) type = 'any[]';
+          if (reference) {
+            type = relation.endsWith('one') ? `${reference}` : `${reference}[]`;
```
tshemsedinov, line 257:
```suggestion
            type = reference + relation.endsWith('one') ? '' : '[]';
```

lib/types.js
```diff
@@ -1,10 +1,258 @@
...
+    } else if (def.many) {
+      ref.name = def.many;
+      ref.relation = 'one-to-many';
+    } else if (def.one) {
+      ref.name = def.one;
+      ref.relation = 'one-to-one';
+    } else if (def.type !== this.name) {
+      ref.name = def.type;
+      ref.relation = 'many-to-one';
+    }
```
tshemsedinov, line 83: Maybe we may use collection of functions here?

lib/types.js
```diff
@@ -1,10 +1,258 @@
...
+    schemaRef.references = new Set([
+      ...schemaRef.references,
+      ...schema.references,
+    ]);
+    schemaRef.relations = new Set([
+      ...schemaRef.relations,
+      ...schema.relations,
+    ]);
+
```
tshemsedinov, line 116:
```suggestion
```

lib/types.js
```diff
@@ -1,10 +1,258 @@
...
+          `In ${this.name} "${path}": type of value must be a ${def.value}`,
+        ];
+      }
+    }
+    return [];
+  },
+  toLong(def) {
+    if (def.type) {
+      return def;
+    }
```
tshemsedinov, line 195:
```suggestion
    if (def.type) return def;
```

lib/util.js
```diff
@@ -0,0 +1,51 @@
+'use strict';
```
tshemsedinov, line 1:
```suggestion
'use strict';

```

lib/util.js
```diff
@@ -0,0 +1,51 @@
...
+const formatLength = (length) => {
+  if (typeof length === 'number') return { max: length };
+  if (Array.isArray(length)) return { min: length[0], max: length[1] };
+  return length;
+};
+
+const omit = (key, obj) => {
+  // eslint-disable-next-line no-unused-vars
+  const { [key]: omitted, ...rest } = obj;
+  return rest;
```
tshemsedinov, line 37:
```suggestion
  const { [key]: omitted, ...rest } = obj;
  return omitted, rest;
```

lib/util.js
```diff
@@ -0,0 +1,51 @@
...
+};
+
+const isFirstUpper = (s) => !!s && UPPER_LETTERS.includes(s[0]);
+
+const isFirstLetter = (s) =>
+  (!!s && UPPER_LETTERS.includes(s[0])) || LOWER_LETTERS.includes(s[0]);
```
tshemsedinov, line 43:
```suggestion
const isFirstUpper = (s) => !!s && UPPER_LETTERS.includes(s[0]);

const isFirstLower = (s) => !!s && LOWER_LETTERS.includes(s[0]);

const isFirstLetter = (s) => isFirstUpper(s) || isFirstLower(s);
```

lib/schema.js
georgolden: I thought it can be better to keep here functions, just to avoid `null` and avoid more checks for future development. So it is similar, that I did with `validate` there is no reason to check if `validate === null (!validate)`. So, goal here - to reduce amount of checks and if statements.

lib/util.js
georgolden: The problem with `metautil` ifFirstUpper function works incorrectly with scpecial chars: `isFirstUpper('?string') === true`

I suggest this variant to migrate to `metautil` instead of being part of `metaschema`

lib/schema.js
```diff
@@ -13,219 +19,188 @@ const SCOPE = ['application', 'global', 'local'];
...
-    const key = Object.keys(col)[0];
-    const value = col[key];
-    return Object.assign(def, { type, key, value });
-  } else {
-    return { type };
+const typeOf = (raw) => {
+  const defType = typeof raw;
+  if (defType === 'object') {
+    const isArray = Array.isArray(raw);
+    if (isArray) return 'array';
```
Paul4Olden, line 92:
```suggestion
    if (Array.isArray(raw)) return 'array';
```

lib/schema.js
```diff
@@ -13,219 +19,188 @@ const SCOPE = ['application', 'global', 'local'];
...
   }
```
Paul4Olden, line 93:
```suggestion
  if (Array.isArray(raw)) return 'array';
```

lib/schema.js
georgolden: I am using metarhia eslint config and this parentheses issue, because of lint error

lib/schema.js
```diff
@@ -13,219 +19,188 @@ const SCOPE = ['application', 'global', 'local'];
...
+      defs[key] = rules.fields[key];
+    }
+    return defs;
   }
 
-  static extractSchema(def) {
-    if (def instanceof Schema) return def;
-    if (def.schema instanceof Schema) return def.schema;
-    return null;
+  extractMetadata(meta) {
```
Paul4Olden, line 188:
```suggestion
  applyMetadata(meta) {
```

lib/schema.js
lundibundi: IMO we shouldn't have format/parse/serialize as "default" functions. Because this is optional behavior and with such approach it becomes not clear how to distinguigsh custom `serialize` vs `do nothing`.

lib/schema.js
```diff
@@ -13,119 +19,80 @@ const SCOPE = ['application', 'global', 'local'];
...
-  const keys = Object.keys(def);
-  for (const key of keys) {
-    if (def[key].shorthand) return key;
+const setValidate = (validate) => {
+  if (!validate) return () => ({ valid: true, errors: [] });
+  if (typeof validate !== 'function') {
+    throw new Error('Validate must be a function');
   }
-  return '';
+  return (value, path) => {
```
lundibundi, line 32: This closure is duplicated between Schemas (and therefore uses more memory), wdyt of having a separate function and call .bind instead?

lib/model.js
```diff
@@ -18,9 +19,19 @@ class Model {
...
+        continue;
+      }
+      const schema = new Schema(name, entity, [this]);
+      this.entities.set(name, schema);
+    }
+    for (const [name, entity] of last) {
       const schema = new Schema(name, entity, [this]);
       this.entities.set(name, schema);
     }
+    last.clear();
```
lundibundi, line 34: This introduces a Model loading race condition, why not put `last` inside of `Model` as a separate internal field?

lib/schema.js
```diff
@@ -135,97 +102,100 @@ class Schema {
...
-      this.references.add(type);
-      if (isFirstUpper(type)) {
-        this.relations.add({ to: type, type: 'many-to-one' });
-      }
-      this.fields[key] = def;
+      if (this.preprocessIndex(key, entry)) continue;
+      const [k, req] = formatRequired.key(key);
+      const { name, def } = this.preprocessType(entry);
+      const type = this.findType(name);
+      if (!type) throw new Error(`Unknown type ${type}`);
```
lundibundi, line 135: We can probably throw in `findType`, wdyt?

lib/util.js
```diff
@@ -0,0 +1,41 @@
+'use strict';
+const UPPER_LETTERS = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ';
+const LOWER_LETTERS = 'abcdefghijklmnopqrstyvwxyz';
+
+const formatRequired = {
+  type: (type, req = true) => {
+    const required = !type.startsWith('?');
+    if (required) return [type, required && req];
+    const name = type.substring(1);
+    return [name, required && req];
```
lundibundi, line 10: IMO it is better to just use objects.

lib/types.js
```diff
@@ -1,10 +1,249 @@
...
+  check(value, path, def) {
+    if (def.enum.includes(value)) return [];
+    return [`Field "${path}" value is not of enum: ${def.enum.join(', ')}`];
+  },
+  toLong(def) {
+    return { type: 'enum', ...def };
+  },
+};
+
+const reference = {
```
lundibundi, line 45: IMO it would be good to separate the per-file and collect in some types/index.js

lib/schema.js
```diff
@@ -135,97 +102,100 @@ class Schema {
...
+      const long = type.toLong(def, Schema, this);
+      this.kind = this.kind || type.kind;
+      return { value: long };
     }
+    this.kind = kind;
+    if (!this.findKind(kind)) throw new Error('Given Kind is invalid');
+    this.applyMetadata(metadata);
+    const extra = kind === 'projection' ? this.projection(metadata) : {};
```
lundibundi, line 124: Nit: We should consider some "streamlined" projection handling to avoid such `if projection else everything else` checks

georgolden:
Closed for https://github.com/metarhia/metaschema/pull/393 here is better implementation and more features

Issue #378, opened 2022-01-11
Array of object support


georgolden:
**Is your feature request related to a problem? Please describe.**
Sometimes I am working with multiple instances of a model like this array of MarketOrder.
So it will be nice to have this syntax supported:
```
{
  array: 'MarketOrder',
}

{
  type: 'array',
  value: {
    type: 'MarketOrder',
  }
}

MarketOrder: {
  price: 'number',
  volume: 'number',
  operation: { enum: ['buy', 'sell'] },
}

{
  array: {
    price: 'number',
    volume: 'number',
    operation: { enum: ['buy', 'sell'] },
  }
}
```

**Describe the solution you'd like**
Support of syntax like this with both Model and Schema in metaschema

**Describe alternatives you've considered**
I think there are no alternatives for this feature.

georgolden:
Done with https://github.com/metarhia/metaschema/pull/400

PR #393, opened 2022-05-10
Rewrite schemas and implement custom types

georgolden:
- [ ] description of changes is added in CHANGELOG.md
- [ ] update .d.ts typings

Need help:
 - or ideas for Preprocessor - improve code quality
 - overall - better naming
 - Model change reorder to remove hardcoded projection

Non breaking:
 - elegant syntax/format for custom types and internal types
 - custom types support
 - preprocessor to reduce Schema preprocess complexity
 - any kind of nested arrays, array of references https://github.com/metarhia/metaschema/issues/378
 - nested schemas with Schema instances support
 - any level of complex nested types support
 - custom validate for field
 - changed validate for schema, simplified for user
 - all user's validate functions support 4 types of syntax: boolean return, throw Error, error message as string return and array of error messages
 - reserved words permitted in schema if kind provided
 - function fields stored in schema
 - shorthand required key for collections support `{ 'array?': 'string' }`
 - many relation now checks in runtime
 - model now loads projections at the end, to fix bug
 - schema kinds moved to separate file and kinds now have logic to remove hardcoded projection from Schema
 - ts interfaces from schema now have relation ids as well
 - deps update metautil
 - nested object support https://github.com/metarhia/metaschema/issues/395
 - syntax for validate function changed

Breaking:
 - Model no more require metavm for browser compatibility: need `impress` update
 - loadModel moved to loader: need `impress` update
 - Model no more writes d.ts file to disk moved to loader as well: need `metasql` update
 - syntax for pg types changed a bit, no more not-working types: need `metasql` update
 - not supporting custom Schema kinds: need to not use custom kinds in schemas folder: `application/schemas/<YourSchema.js>`

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -9,11 +9,22 @@ class Model {
     this.entities = new Map();
     this.database = database;
     this.order = new Set();
+    this.last = new Set();
     this.warnings = [];
     for (const [name, entity] of entities) {
+      const first = Object.keys(entity)[0];
+      if (first === 'Projection') {
+        this.last.add([name, entity]);
+        continue;
+      }
       const schema = new Schema(name, entity, [this]);
       this.entities.set(name, schema);
     }
+    for (const [name, entity] of this.last) {
+      const schema = new Schema(name, entity, [this]);
+      this.entities.set(name, schema);
+    }
+    this.last.clear();
     this.preprocess();
   }
 
--- a/lib/model.js
+++ b/lib/model.js
@@ -25,7 +25,6 @@ class Model {
       const schema = new Schema(name, entity, [this]);
       this.entities.set(name, schema);
     }
-    last.clear();
     this.preprocess();
   }
 
--- a/lib/model.js
+++ b/lib/model.js
@@ -2,6 +2,7 @@
 
 const { Schema } = require('./schema.js');
 const { prepareTypes } = require('./types.js');
+const { firstKey } = require('./util.js');
 
 class Model {
   constructor(types, entities, database = null) {
@@ -12,7 +13,7 @@ class Model {
     this.warnings = [];
     const last = new Set();
     for (const [name, entity] of entities) {
-      const first = Object.keys(entity)[0];
+      const first = firstKey(entity);
       if (first === 'Projection') {
         last.add([name, entity]);
         continue;
--- a/lib/model.js
+++ b/lib/model.js
@@ -1,37 +1,20 @@
 'use strict';
 
-const fsp = require('fs').promises;
 const { Schema } = require('./schema.js');
-const { readDirectory } = require('./loader.js');
-const { TYPES } = require('./types.js');
+const { prepareTypes } = require('./types.js');
 
 class Model {
   constructor(types, entities, database = null) {
-    this.types = types;
+    this.types = prepareTypes(types);
     this.entities = new Map();
     this.database = database;
     this.order = new Set();
     this.warnings = [];
     for (const [name, entity] of entities) {
-      if (entity.type) {
-        this.warnings.push(
-          `Warning: 'type' property is restricted in database schemas: ${name}`
-        );
-      }
       const schema = new Schema(name, entity, [this]);
       this.entities.set(name, schema);
     }
     this.preprocess();
-  }
-
-  static async load(modelPath, systemTypes = {}) {
-    const structs = await readDirectory(modelPath);
-    const database = structs.get('.database') || null;
-    const customTypes = structs.get('.types') || {};
-    structs.delete('.database');
-    structs.delete('.types');
-    const types = { ...TYPES, ...systemTypes, ...customTypes };
-    return new Model(types, structs, database);
   }
 
   preprocess() {
@@ -52,7 +35,8 @@ class Model {
   reorderEntity(name, base) {
     const entity = this.entities.get(name);
     if (!entity) return;
-    for (const ref of entity.references) {
+    const { references } = entity;
+    for (const ref of references) {
       if (ref === name) continue;
       if (ref === base) {
         this.warnings.push(`Recursive dependency: ${name}.${base}`);
@@ -63,14 +47,14 @@ class Model {
     this.order.add(name);
   }
 
-  async saveTypes(outputFile) {
+  get dts() {
     const { entities, order } = this;
     const dts = [];
     for (const name of order) {
       const schema = entities.get(name);
       dts.push(schema.toInterface(name));
     }
-    await fsp.writeFile(outputFile, dts.join('\n\n') + '\n');
+    return dts.join('\n\n') + '\n';
   }
 }
 
--- a/lib/model.js
+++ b/lib/model.js
@@ -46,8 +46,7 @@ class Model {
   reorderEntity(name, base) {
     const entity = this.entities.get(name);
     if (!entity) return;
-    const { references } = entity;
-    for (const ref of references) {
+    for (const ref of entity.references) {
       if (ref === name) continue;
       if (ref === base) {
         this.warnings.push(`Recursive dependency: ${name}.${base}`);
--- a/lib/model.js
+++ b/lib/model.js
@@ -9,22 +9,22 @@ class Model {
     this.entities = new Map();
     this.database = database;
     this.order = new Set();
-    this.last = new Set();
     this.warnings = [];
+    const last = new Set();
     for (const [name, entity] of entities) {
       const first = Object.keys(entity)[0];
       if (first === 'Projection') {
-        this.last.add([name, entity]);
+        last.add([name, entity]);
         continue;
       }
       const schema = new Schema(name, entity, [this]);
       this.entities.set(name, schema);
     }
-    for (const [name, entity] of this.last) {
+    for (const [name, entity] of last) {
       const schema = new Schema(name, entity, [this]);
       this.entities.set(name, schema);
     }
-    this.last.clear();
+    last.clear();
     this.preprocess();
   }
 
--- a/lib/preprocessor.js
+++ b/lib/preprocessor.js
@@ -1,0 +1,59 @@
+'use strict';
+
+const { isFirstUpper, toLowerCamel } = require('metautil');
+const { formatters, firstKey, omit } = require('./util.js');
+
+class Preprocessor {
+  constructor(types, Schema) {
+    this.Schema = Schema;
+    this.types = types;
+  }
+
+  parse(source) {
+    const srcType = Array.isArray(source) ? 'tuple' : typeof source;
+    return this.recognize(srcType, source);
+  }
+
+  single(def) {
+    const { types } = this;
+    const [type, required] = formatters.type(
+      def.type || def,
+      def.required || true
+    );
+    if (isFirstUpper(type)) {
+      return new types.reference({ one: type, required, ...def });
+    }
+    if (!types[type]) throw new Error(`Unknown type ${type}`);
+    return new types[type]({ required, ...def }, this);
+  }
+
+  recognize(srcType, source) {
+    const { types } = this;
+    if (srcType === 'string') {
+      return this.single(source);
+    }
+    if (srcType === 'object') {
+      const first = firstKey(source);
+      // schema with kind
+      if (isFirstUpper(first)) {
+        const [metadata, defs] = omit(first, source);
+        metadata.kind = toLowerCamel(first);
+        return { defs, metadata };
+      }
+      // longform definition
+      if (source.type) return this.single(source);
+      // shorthand
+      const [type, required] = formatters.key(first, source.required);
+      if (types[type]) {
+        const [def, rest] = omit(first, source);
+        return new types[type]({ [type]: def, required, ...rest }, this);
+      }
+      // schema
+      return { defs: source };
+    }
+    if (srcType === 'function') return { skip: source };
+    throw new Error(`Invalid definition: "${source}" of type ${srcType}`);
+  }
+}
+
+module.exports = { Preprocessor };
--- a/lib/preprocessor.js
+++ b/lib/preprocessor.js
@@ -4,9 +4,11 @@ const { isFirstUpper, toLowerCamel } = require('metautil');
 const { formatters, firstKey, omit } = require('./util.js');
 
 class Preprocessor {
-  constructor(types, Schema) {
+  constructor(root, Schema) {
     this.Schema = Schema;
-    this.types = types;
+    this.root = root;
+    this.kinds = Schema.KIND;
+    this.types = root.getTypes();
   }
 
   parse(source) {
@@ -16,10 +18,7 @@ class Preprocessor {
 
   single(def) {
     const { types } = this;
-    const [type, required] = formatters.type(
-      def.type || def,
-      def.required || true
-    );
+    const [type, required] = formatters.type(def.type, def.required);
     if (isFirstUpper(type)) {
       return new types.reference({ one: type, required, ...def });
     }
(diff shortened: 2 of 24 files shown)


lib/prototypes/reference.js
```diff
@@ -0,0 +1,26 @@
...
+  format(def) {
+    const reference = def.one || def.many;
+    const relation = def.many ? 'many-to-one' : 'one-to-many';
+    this.type = reference;
+    this.references.add(reference);
+    this.relations.add({ to: reference, type: relation });
+  },
+};
+
+module.exports = { reference, many: reference, one: reference };
```
tshemsedinov, line 26: Why do we need to export same object 3 times?

lib/prototypes/collections.js
```diff
@@ -0,0 +1,91 @@
...
+
+const set = {
+  ...array,
+  type: 'set',
+  isInstance(value) {
+    return Reflect.getPrototypeOf(value) === Set.prototype;
+  },
+};
+
+module.exports = { object, map, array, set };
```
tshemsedinov, line 91: Maybe it's better to hold each type in a separate file like `({ /* JSON5 metadata */ })` and load it with metavm

test/database.js
```diff
@@ -3,9 +3,9 @@
 const metatests = require('metatests');
 const { Schema } = require('..');
 
-metatests.test('Schema: database', (test) => {
+metatests.test('Database: schema Registry', (test) => {
   const raw = {
-    Registry: {},
+    Registry: { scope: 'application', store: 'persistent', allow: 'write' },
```
tshemsedinov, line 8: Those fields are expected to be default

lib/prototypes/reference.js
georgolden: It will be included in Schema and checked as well. In previous version no fields like this were checked

georgolden: I made improvement. Now moved to prototypes

georgolden:
@tshemsedinov https://github.com/metarhia/metaschema/issues/395 Done

lib/schema.js
```diff
@@ -1,214 +1,111 @@
...
+const RESERVED = ['validate', 'parse', 'serialize', 'format'];
 
 class Schema {
-  constructor(name, raw, namespaces) {
-    const short = shorthand(raw) || raw.type;
-    const defs = short ? { value: raw } : raw;
-    this.name = name;
-    this.kind = short ? 'scalar' : 'struct';
+  constructor(name, raw, namespaces = []) {
+    Schema.KIND = KIND;
```
lundibundi, line 22: Why is this here and not along with other static properties?

lib/schema.js
```diff
@@ -337,7 +217,6 @@ class Schema {
   }
 }
 
-Schema.KIND = KIND;
 Schema.KIND_STORED = KIND_STORED;
```
lundibundi, line 220: Minimum supported Node.js version by metaschema is 12.9 so we can just use `static` properties. https://node.green/#ES2022-features-static-class-fields-public-static-class-fields

lib/util.js
```diff
@@ -0,0 +1,59 @@
...
+    }
+  },
+};
+
+const harmonize = (result) => {
+  if (typeof result === 'boolean') {
+    return result ? '' : ['Validation error'];
+  }
+  if (result) return Array.isArray(result) ? result : [result];
+};
```
lundibundi, line 41:
```suggestion
const harmonizeErrors = (result) => {
  if (typeof result === 'boolean') {
    return result ? null : ['Validation error'];
  }
  if (result) return Array.isArray(result) ? result : [result];
  return null;
};
```
wdyt?

lib/schema.js
```diff
@@ -1,214 +1,111 @@
...
-        const parent = this.findReference(this.parent);
-        for (const key of metadata.fields) {
-          defs[key] = parent.fields[key];
-          keys.push(key);
-        }
-      }
+  preprocess(raw, prep) {
+    const source = prep.parse(raw);
+    if (isType(source)) {
+      this.fields = source;
```
lundibundi, line 56: Does this mean that `fields` is not longer just an object with fields? This would probably break some things, why do we need this?

lib/prototypes/abstract.js
```diff
@@ -0,0 +1,52 @@
...
+  checks: {},
+  check(value, path, root) {
+    try {
+      const typeErr = harmonize(this.checkType(value, path, root));
+      if (typeErr) return typeErr;
+      if (this.validate) {
+        const validErr = harmonize(this.validate(value, path, root));
+        if (validErr) return validErr;
+      }
+      for (const [name, subCheck] of Object.entries(this.checks)) {
```
lundibundi, line 16: Why not have `checks` as Map or even an Array then?

lib/prototypes/abstract.js
```diff
@@ -0,0 +1,52 @@
...
+        if (this[name]) subCheck(value, path, this);
```
lundibundi, line 17: What if one of the `subCheck` wants to return an error, shouldn't we collect these into an array and return it?

lib/prototypes/abstract.js
```diff
@@ -0,0 +1,52 @@
...
+    const { formatters } = this;
+    for (const key of Object.keys(def)) {
+      if (key === 'type' || key === this.type) continue;
+      if (formatters[key]) this[key] = formatters[key](def[key]);
+      else this[key] = def[key];
+    }
+    this.construct(def, preprocessor);
+    const { type } = this;
+    this.type = type;
+    this.references.add(type);
```
lundibundi, line 48:
```suggestion
    this.references.add(this.type);
```
?

lib/prototypes/collections.js
```diff
@@ -0,0 +1,100 @@
...
+  entries(value) {
+    return Object.entries(value);
+  },
+};
+
+const map = {
+  ...object,
+  type: 'map',
+  isInstance(value) {
+    return Reflect.getPrototypeOf(value) === Map.prototype;
```
lundibundi, line 53: This would probably not work in metavm contexts and same goes for other such checks, since they prototypes from different sandboes will probably not be reference-equal. Have you verifies this case? (`schema` created in one sandbox and `value` created in another)

lib/prototypes/collections.js
```diff
@@ -0,0 +1,100 @@
...
+    const source = def.type ? def.value : def[type];
+    const child = prep.parse(source);
+    if (isType(child)) {
+      this.value = child;
+      this.metadataFrom(child);
+    } else {
+      const schema = new prep.types.schema(child.defs, prep);
+      this.references.add('schema');
+      this.metadataFrom(schema);
+      this.value = schema;
```
lundibundi, line 84: why are we storing schema under `value` field?

lib/schema.js
```diff
@@ -233,15 +130,24 @@ class Schema {
...
+    const type = DEFAULT[name];
+    if (type) return type;
     return null;
   }
 
+  getTypes() {
+    let types = {};
+    for (const ns of this.namespaces) {
+      if (ns.types) types = { ...types, ...ns.types };
+    }
```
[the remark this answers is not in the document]
georgolden: will be referenced in a commit. Thx a lot

lib/schema.js
```diff
@@ -233,15 +130,24 @@ class Schema {
...
+    if (Object.keys(types).length === 0) types = DEFAULT;
+    return types;
```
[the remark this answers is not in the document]
georgolden: will be referenced in a commit. Thx a lot

lib/schema.js
georgolden: The reason for this is outdated `"eslint-config-metarhia"`. It depends on old eslint, so I got eslint error, which can not be disabled easily without update. I see no big issue with this approach. Syntax is OK. But we need to update our linter to support new ES features

lib/util.js
georgolden:
```js
const harmonize = (error) => {
  if (typeof error === 'boolean') {
    return error ? null : ['Validation error'];
  }
  if (error) return Array.isArray(error) ? error : [error];
  return null;
};
```
wdyt?

lib/schema.js
georgolden: In case if schema is a single type, we don't need to keep useless field `{ value: { type: 'string' } }` we can store any type shorthand directly in fields and check by constructor name with `isType`

georgolden: Logs look like this:
```js
fields: Type {
    references: Set(1) { 'string' },
    relations: Set(0) {},
    required: true,
    type: 'string'
  },
```

lib/util.js
georgolden: Merged your changes and kept short name

lib/prototypes/abstract.js
georgolden: It never will be large, so no reason to use map. But I will try an array

lib/prototypes/abstract.js
georgolden: I did not use harmonize for subCheck. And I do not expect to run all checks, so I will add this option, cause we have similar behavior in Schema. Thx

lib/prototypes/abstract.js
georgolden: this needed for type to be visible in logs, otherwise type will be always in prototype and will be not visible in logs. And for reference it is a difference. Diff:
```js
apartment: Type {
      references: [Set],
      relations: Set(0) {},
      required: true,
      type: 'string'
    },
```
```js
persons: Type {
      references: [Set],
      relations: [Set],
      many: 'Person',
      required: true,
      type: 'Person'
    }
```

georgolden: Without this lines:
```js
apartment: Type {
      references: [Set],
      relations: Set(0) {},
      required: true
    },
```
```js
persons: Type {
      references: [Set],
      relations: [Set],
      many: 'Person',
      required: true,
      type: 'Person'
    }
```

lib/prototypes/collections.js
georgolden: This is from old version. Replaced with checking constructor name

lib/prototypes/collections.js
georgolden: schema is only invariant, that keeps exist. So it requires quite complex explanation to explain, why we can't avoid schema as an invariant. In short: array defenition has a reserved field called value, to describe type of each element in array.
```js
{ type: 'array', value: 'string' }
{ type: 'array', value: { name: 'string' } }
```

lundibundi: So this is for `array` only, right?

georgolden: for array and set - `{ value: Type/Schema }`
for object and map - `{ key: Type, value: Type/Schema }`
so this mean for all collections to support any nested type or nested schemas

lib/prototypes/abstract.js
georgolden: Changed to an array

lib/prototypes/abstract.js
georgolden: Changed to an array

tshemsedinov (approved):
I'd prefer to remove comments but it's ok to do it in next PR after naming improving.

lib/model.js
```diff
@@ -1,39 +1,34 @@
...
+        continue;
       }
       const schema = new Schema(name, entity, [this]);
       this.entities.set(name, schema);
     }
+    for (const [name, entity] of last) {
+      const schema = new Schema(name, entity, [this]);
+      this.entities.set(name, schema);
+    }
+    last.clear();
```
lundibundi, line 28: Nit:
```suggestion
```

lib/model.js
```diff
@@ -52,7 +47,8 @@ class Model {
   reorderEntity(name, base) {
     const entity = this.entities.get(name);
     if (!entity) return;
-    for (const ref of entity.references) {
+    const { references } = entity;
+    for (const ref of references) {
```
lundibundi, line 51: Nit: destructiring is redundant here.

lib/model.js
```diff
@@ -1,39 +1,33 @@
...
+const { prepareTypes } = require('./types.js');
 
 class Model {
   constructor(types, entities, database = null) {
-    this.types = types;
+    this.types = prepareTypes(types);
     this.entities = new Map();
     this.database = database;
     this.order = new Set();
+    this.last = new Set();
```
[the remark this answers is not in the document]
georgolden: Done

lib/preprocessor.js
```diff
@@ -0,0 +1,64 @@
...
+    // scalar or refernce shorthand
+    if (srcType === 'string') {
+      return this.single({ type: source });
+    }
+    if (srcType === 'object') {
+      const first = firstKey(source);
+      // schema with kind
+      if (isFirstUpper(first)) {
+        const { kinds, root } = this;
+        const [meta, fields] = omit(first, source);
```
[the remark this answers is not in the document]
georgolden: Done

lib/schema.js
```diff
@@ -1,214 +1,111 @@
...
+    const { index, primary, unique, many } = defs;
+    const isIndex = Array.isArray(index || primary || unique);
+    if (isIndex || many) this.indexes[key] = defs;
     return isIndex;
   }
 
-  static from(raw, namespaces) {
-    return new Schema('', raw, namespaces);
+  updateMetadata(meta) {
+    this.references = new Set([...this.references, ...meta.references]);
```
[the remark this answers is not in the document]
georgolden: Done

lib/schema.js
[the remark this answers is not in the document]
georgolden: Done

lib/schema.js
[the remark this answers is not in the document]
georgolden: Done

lib/schema.js
```diff
@@ -1,214 +1,111 @@
...
-          { required },
-          entry.schema ? entry : {},
-          { type: 'schema', schema }
-        );
-        this.references = new Set([...this.references, ...schema.references]);
-        this.relations = new Set([...this.relations, ...schema.relations]);
+      if (this.preprocessIndex(key, entry)) continue;
+      const child = prep.parse(entry);
+      if (child.skip) {
+        this.fields[key] = child.skip;
```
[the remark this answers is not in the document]
georgolden: Done

lib/schema.js
georgolden: Done

lib/schema.js
georgolden: Can be done after eslint config update

lib/util.js
georgolden: Done

lib/schema.js
georgolden: Answered

lib/prototypes/abstract.js
georgolden: Impossible to use array, because field may not use `subCheck` example: `{ type: 'string' }` may not use length at all. `{ type: 'string', length: 10 }` may use length `subCheck`

georgolden: Answered

lib/prototypes/abstract.js
georgolden: Here we need to check if there is subCheck rule in type definition. Can not be changed.

georgolden: Answered

lib/prototypes/abstract.js
georgolden: Answered

lib/prototypes/collections.js
georgolden: Done

lib/prototypes/collections.js
georgolden: Answered

lib/model.js
georgolden: Done

tshemsedinov:
Landed in https://github.com/metarhia/metaschema/commit/ee1bca5e7512ede93b0dfdc9d534eef0f03d14a2

PR #457, opened 2023-09-16
Optimize kinds.js module. Add tests

rustworthy:
- [ ] description of changes is added in CHANGELOG.md // N/A
- [ ] update .d.ts typings // N/A

Changes:

- return early if schema name OR (!) fields not specified when building projection;
- check that 'root' passed to "projection" func respects Schema API;
- add tests for "projection" func;
- treat projection as a private case  of 'kindMemory' and simplify 'getKindMetadata' dispatcher func;
- use 'KIND_STORED_DEFAULT_META' and 'KIND_MEMORY_DEFAULT_META' constants to store module defaults;

As a side-note:
both 'schema' and 'fields' are words kind of reserved for Schema instance and field attribute of schema instance accordingly,
so I would suggest using schemaName and fieldNames where appropriate; not always this is clear from the context.

### DIFF
--- a/lib/kinds.js
+++ b/lib/kinds.js
@@ -12,6 +12,25 @@ const SCOPE = ['application', 'global', 'local'];
 const STORE = ['persistent', 'memory'];
 const ALLOW = ['write', 'append', 'read'];
 
+const KIND_STORED_DEFAULT_META = {
+  scope: 'application',
+  store: 'persistent',
+  allow: 'write',
+};
+const KIND_MEMORY_DEFAULT_META = {
+  scope: 'local',
+  store: 'memory',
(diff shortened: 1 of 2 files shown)


lib/kinds.js
```diff
@@ -12,38 +12,61 @@ const SCOPE = ['application', 'global', 'local'];
 const STORE = ['persistent', 'memory'];
 const ALLOW = ['write', 'append', 'read'];
 
+const KIND_STORED_DEFAULT_META = {
+  scope: 'application',
+  store: 'persistent',
+  allow: 'write',
+};
```
tshemsedinov, line 19:
```suggestion
};

```

